PR #17, opened 2026-07-13
fix(skill_triggered): detect skill engagement by regex, not the Skill tool

Mihaiii:
Neither Antigravity CLI nor Codex CLI has a Skill tool — only Claude Code does. Both engage a skill by reading its SKILL.md off disk: Antigravity via its file-view/read tool (view_file -> Read), Codex via shell (cat/sed, recorded as Bash).

Detection now matches a skill-specific regex over the serialized command parameters instead of inspecting cmd.tool_name == "Skill".

This scores identically across all agents and requires no Skill tool.

Add tests for a Read tool whose path ends in SKILL.md (Antigravity's .agents/skills/ layout) and a guard that a SKILL.md outside any skill dir does not false-positive.


Mihaiii:
Closing as redundant.

PR #18, opened 2026-07-13
chore(evalboard): address npm Dependabot alerts

bai-uipath:
## What

Clears the open npm Dependabot alerts on the evalboard frontend and stops them from piling up again. All 26 alerts came from this one Next.js app; the Python core has none.

Three parts:

- Adds an npm version-updates block to `dependabot.yml` (grouped, monthly) and batches the existing github-actions bumps into one grouped PR too. Until now nothing routinely refreshed the frontend, so it drifted until a CVE fired — this keeps next/React/Azure ahead of most advisories going forward, as a single low-noise PR per ecosystem.

- Patches the runtime alerts: `next` to 15.5.20, plus `postcss` / `uuid` / `fast-xml-builder` / `@babel/core` pulled to their fixed versions.

- Lazy-loads the Azure SDK and moves `@azure/*` to `optionalDependencies`, so local-mode readers (and OSS users who skip optional installs) never pull it. `@azure/storage-blob` is capped below 12.33.0, which raised its floor to node 22, to stay on the project's node 20 baseline.

## Why

Public-repo hygiene on a repo we want to showcase. The frontend runs only as a localhost dev tool or a read-only, SSO-gated internal dashboard, so real exploit risk is low — but keeping the alert list clean and self-maintaining is worth doing.

## Testing

- **Local build + tests:** `pnpm run verify` (typecheck + 324 unit tests + `next build`) passes on Node 20.

- **Azure fetch (remote mode):** exercised the app against Blob Storage — it lists and reads runs through the lazy-loaded Azure SDK via `DefaultAzureCredential`, confirming the async `getContainer` / dynamic-import refactor still fetches from Azure as before.

- **Deploy:** reproduced the hosted deploy's exact install (`npm install --omit=dev` from `package.json`, per `deploy.sh`) and confirmed `@azure/*` (storage-blob 12.32.0) installs and resolves on the deploy's Node 22 — moving Azure to `optionalDependencies` does not break the hosted deploy.

## Related Dependabot settings (not code in this PR)

The dev-only alerts (~half of the list: vitest, vite, esbuild, ws, form-data) never ship — the `output: "standalone"` build excludes dev deps. Rather than bump them, a repo Dependabot auto-triage rule ("Dismiss development-scoped npm alerts", indefinitely) now auto-dismisses dev-scoped npm alerts — clearing those (including the two "critical" vitest alerts) and keeping future dev-only alerts out of the queue.


akshaylive (approved):
🚢

PR #19, opened 2026-07-13
fix(sandbox): exclude home-dir dotfiles from capture_to artifacts

joeysbase:
## Summary

- `_WORKSPACE_CAPTURE_IGNORE` was missing common Linux home-directory entries (`.cache`, `.config`, `.npm`, `.local`, shell dotfiles)
- When `working_dir: /root` (or any HOME-overlapping path), tools like uv/pip/npm write these into the WORKDIR — and `capture_to` was copying all of them verbatim into the artifacts directory
- Extends the ignore list so only actual task deliverables land in artifacts

## Reproduction

Observed with the `skillsbench-court-form-filling` task (uses `working_dir: /root`): artifacts contained `.cache/uv/`, `.config/uv/`, `.config/coder-eval/`, `.npm/`, `.bashrc`, `.profile`, `.wget-hsts` alongside the intended `sc100-filled.pdf`.

## Test plan

- [x] Run a task with `sandbox.docker.working_dir: /root` and confirm artifacts only contain task deliverables
- [x] Confirm existing tasks with non-HOME workdirs are unaffected

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ dependencies = [
     "pydantic-settings>=2.14.2",
     "pyyaml>=6.0.3",
     "typer>=0.24.1",
+    "click>=8.3.3",
     "rich>=14.3.3",
     "python-dotenv>=1.2.2",
     "anthropic>=0.86.0",
--- a/src/coder_eval/sandbox.py
+++ b/src/coder_eval/sandbox.py
@@ -27,7 +27,25 @@ logger = logging.getLogger(__name__)
 # sandbox-created bulk. `.claude` is the RW lean copy of ~/.claude (carries
 # .credentials.json). The rest are sandbox infra the standard artifacts path
 # also drops. Matched by basename at every level (shutil.ignore_patterns).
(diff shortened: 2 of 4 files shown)


bai-uipath (approved):
**Fix what you agree with, otherwise lgtm.** Correct, focused fix with a narrow blast radius (docker WORKDIR mode only, and the capture runs post-grading so no verdict can change). Denylist is the right shape here — the noise-vs-deliverable distinction isn't mechanically derivable and isn't task-specific, so a curated constant is exactly what should encode it. One real ask before merge (expand it), the rest optional.

- **Expand the denylist to the credential family — before merge.** The reason `.claude` is excluded is credential-leak prevention, and these captured artifacts get uploaded — but `.aws`, `.ssh`, `.azure`, `.netrc`, `.gitconfig`, `.docker` aren't covered. If any eval image ships those under `/root`, they leak into artifacts the same way `.credentials.json` would have. This PR is the natural place to close that class. Fix: add them to the same tuple in `sandbox.py:30` (worth a quick check of what the images actually put under `/root` first), and note in the comment that this is a security denylist, not just noise suppression.

PR #23, opened 2026-07-14
feat(evalboard): show conversation transcript for simulation tasks

CarlesUIPath:
## Summary

Simulation runs already write a clean user<->agent transcript to conversation.log (orchestrator._log_conversation), but the evalboard never surfaced it. Add a Conversation section to the task detail page that renders the dialog as chat bubbles, below the Success criteria section. Gated on parsed turns, so single-shot tasks (which never write conversation.log) are unaffected.

## Changes

- lib/runs.ts: readConversationLog (mirror of readLogTail) + a conversation parser returning ConversationTurn[].
- _sections.tsx: ConversationSection renders turns as bubbles (user right/gray, agent left/blue). 
- page.tsx: fetch + parse, render below Success criteria. It only renders the conversation if "conversation.log" exists in the runs/<run-id> directory.

## Tests

Tests mirror repo conventions: unit tests for parseConversation (alongside parseMessages) and a render test for ConversationSection (alongside message-timeline). readConversationLog is left untested, matching its untested twin readLogTail.

## Ui Example

When running a simulation task, a "conversation.log" text is generated. This is rendered in the run page after the Success criteria" as follows.
image

### DIFF
--- a/evalboard/app/runs/[id]/[...task]/__tests__/conversation.test.tsx
+++ b/evalboard/app/runs/[id]/[...task]/__tests__/conversation.test.tsx
@@ -1,0 +1,40 @@
+import { describe, expect, test } from "vitest";
+import { render, screen } from "@testing-library/react";
+import type { ConversationTurn } from "@/lib/runs";
+import { ConversationSection } from "../_sections";
+
+function makeTurn(overrides: Partial<ConversationTurn> = {}): ConversationTurn {
+    return {
+        role: "USER",
+        turn: 1,
+        metadata: null,
+        text: "hello",
+        ...overrides,
+    };
+}
+
+describe("ConversationSection", () => {
+    test("renders both turns' text and roles", () => {
+        render(
+            <ConversationSection
+                turns={[
+                    makeTurn({ role: "USER", text: "please write fizzbuzz" }),
+                    makeTurn({ role: "AGENT", text: "here is the code" }),
+                ]}
+            />,
+        );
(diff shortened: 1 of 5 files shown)


bai-uipath (approved):
**Approve.** Clean, well-scoped feature — parser, presentation, and page wiring live in the right seams, and it's verified against the actual writer. All the design calls here are the ones I'd make.

- **Parser matches the writer exactly.** `parseConversation` targets the format `_log_conversation` emits (orchestrator.py:1843) — header shape, the ` — ` metadata separator (both U+2014), and the `header\n{body}\n\n` framing all line up. The always-present opener metadata (`pinned initial_prompt` / `simulator-generated opener`) parses correctly.
- **The no-simulation default is right, and better than described.** Gating on parsed-turns (`conversation.length > 0`) rather than file existence means single-shot tasks render nothing, and empty-file / no-header-garbage / read-error cases all collapse to `[]` and hide cleanly — no crash, no empty section.
- **Dialogue-only rendering is the correct split.** Tool-use and thinking are stripped on the Python side before the log is written, so bubbles show the clean final utterance; the per-turn caption surfaces `N tool calls` + cost, and the full trajectory already lives in the message-timeline / CostExplorerSection below. Not redundant with the timeline.
- **Tests mirror repo conventions.** Parser unit tests beside `parseMessages`, render test beside `message-timeline`; two-turn / empty / metadata / multi-line all covered. Appropriately scoped.

rockymadden (approved):

PR #26, opened 2026-07-16
fix(codex): keep mock CLIs shadowed in bash login shells

dmorosanu:
## Problem

Codex issues every shell command through the user's default shell as a login shell (`bash -lc` on Linux, `zsh -lc` on macOS). The login shell re-sources the system profile chain (`/etc/profile` on Linux, `/etc/zprofile`'s `path_helper` on macOS), which resets PATH and silently drops the `env_path_prepend` mock-CLI prepend - so `uip` resolves to the real CLI and mocked eval tasks investigate the real tenant instead of the fixtures. In the 2026-07-15 gpt-5.4 weekly run, 5 of 16 uipath-troubleshoot failures were this contamination (transcripts cite real-tenant jobs absent from all fixtures, while the mocked evidence was never served).

## Fix

When mocks are configured (POSIX only), `CodexAgent.start()` generates a per-task HOME containing profile files that:

1. export the ORIGINAL home back (the shell uses env HOME only to pick the profile file; git config, tool caches, and `$HOME`-relative sourcing keep working),
2. source the original home's own profile chain,
3. re-prepend the mock dirs to PATH - last word after any reset.

Coverage per shell: `.bash_profile`/`.profile` for bash/sh (selected via env `HOME`), and `.zshenv`/`.zprofile`/`.zshrc` for zsh - macOS's default shell - selected via env `ZDOTDIR` (`.zshrc` also feeds codex's shell snapshot, which sources it explicitly; `.zshenv` re-pins `ZDOTDIR` in case the user's own `.zshenv` redefines it).

The app-server env points `HOME` and `ZDOTDIR` at the generated dir and pins `CODEX_HOME` to the real codex state dir (auth and sub-agent rollout recovery unaffected). Cleaned up on `stop()`/`kill_sync()`, exception-safe, rerunnable across `start()` retries.

Known residual gap: a nested bash/sh login shell inside a command re-reads the real profiles and loses the prepend again (documented in the docstring). Nested zsh keeps it - `ZDOTDIR` stays exported.

Windows needs none of this: codex there shells through `powershell -NoProfile -Command` or `cmd /c`, neither of which re-sources a profile chain that resets PATH, so the plain env prepend survives - hence the POSIX-only gate.

## Validation

- Real `bash -lc` + the real skills-repo mock dispatcher in WSL Ubuntu 24.04: pre-fix resolves the real CLI, post-fix returns the scenario's manifest fixture.
- Hermetic end-to-end tests with real shells: bash (`bash -lc`) and zsh (`zsh -lc`, `zsh -lic` snapshot shape, nested `zsh -lc`) - mock dir first on PATH, `$HOME` restored, original dotfiles sourced - even when every original profile resets PATH.
- 3413/3413 tests pass on Linux (WSL), 92 + 2 posix-skips on Windows; ruff + pyright clean; no regressions in the full suite.

### DIFF
--- a/src/coder_eval/agents/codex_agent.py
+++ b/src/coder_eval/agents/codex_agent.py
@@ -5,7 +5,9 @@ import contextlib
 import json
 import logging
 import os
+import shlex
 import shutil
+import tempfile
 import time
 from collections.abc import Callable
 from datetime import datetime
@@ -214,6 +216,11 @@ def _message_uncached_input(m: AssistantMessage) -> int:
 # supports the Responses API (it rejects `wire_api = "chat"` as "no longer
 # supported"), so this is a fixed constant, not an operator knob.
 _CODEX_WIRE_API = "responses"
+
+# Login-shell profile files generated into the per-task HOME (see
+# _setup_login_shell_home). ``.bash_profile`` is what ``bash -l`` reads;
+# ``.profile`` covers ``sh``/``dash`` login shells.
+_LOGIN_PROFILE_NAMES = (".bash_profile", ".profile")
 
 
 def _get_item_root(notification: Any) -> Any:
@@ -649,6 +656,7 @@ class CodexAgent(Agent[CodexAgentConfig]):
         self.thread: Any = None
         self.working_directory: Path | None = None
         self._env_path_prepend: list[str] = []
+        self._login_shell_home: Path | None = None
         # _state / _iteration / _iteration_was_incremented / pending_turn lifecycle
         # bookkeeping lives on the Agent base class (shared defaults + helpers).
         self._log = PrefixedAdapter(logger, {"prefix": instance_name})
@@ -677,6 +685,7 @@ class CodexAgent(Agent[CodexAgentConfig]):
         """
         self.working_directory = Path(working_directory)
         self._env_path_prepend = list(env_path_prepend or [])
+        self._setup_login_shell_home()
         self._state = AgentState.WORKING
 
         try:
@@ -872,6 +881,7 @@ class CodexAgent(Agent[CodexAgentConfig]):
         self._close_client()
         self.thread = None
         self._active_turn_handle = None
+        self._cleanup_login_shell_home()
         self._mark_stopped()
 
     async def kill(self) -> None:
@@ -1071,7 +1081,86 @@ class CodexAgent(Agent[CodexAgentConfig]):
             path_key = next((k for k in os.environ if k.upper() == "PATH"), "PATH")
             env[path_key] = os.pathsep.join([*self._env_path_prepend, os.environ.get(path_key, "")])
             self._log.debug(f"PATH prepend: {os.pathsep.join(self._env_path_prepend)}")
+        if self._login_shell_home is not None:
+            # Point login shells at the generated profile dir (see
+            # _setup_login_shell_home) while pinning codex state (auth, rollout
+            # sessions) to its real location — _codex_home() reads the same
+            # resolution for sub-agent rollout recovery, so both sides agree.
+            env["HOME"] = str(self._login_shell_home)
+            env["CODEX_HOME"] = str(self._codex_home())
         return env if env else None
+
+    @staticmethod
+    def _login_shell_profiles_supported() -> bool:
+        """Whether to generate login-shell profile shims (POSIX shells only)."""
+        return os.name == "posix"
+
+    def _setup_login_shell_home(self) -> None:
+        """Create a per-task HOME whose profiles restore the mock PATH prepend.
+
+        Codex issues every shell command as ``bash -lc``. A login shell
+        re-sources ``/etc/profile``, which unconditionally RESETS PATH —
+        silently dropping the mock-CLI prepend passed via the app-server
+        environment, so bare commands resolve to the REAL CLIs (real-tenant
+        contamination). ``~/.bash_profile`` is sourced AFTER ``/etc/profile``,
+        so a generated per-task HOME (wired up in ``_build_codex_env``; codex
+        state stays in CODEX_HOME) gets the last word and re-prepends the mock
+        dirs. Per-task rather than the user's real dotfiles so parallel tasks
+        with different mocks cannot collide. No-op without mock dirs or on
+        non-POSIX hosts (Windows codex does not shell through bash).
+        """
+        self._cleanup_login_shell_home()
+        if not (self._env_path_prepend and self._login_shell_profiles_supported()):
+            return
+        original_home = os.environ.get("HOME", "")
+        # The profile only ever executes under a POSIX shell, so the PATH
+        # separator is ':' regardless of the host building it.
+        quoted_prepend = shlex.quote(":".join(self._env_path_prepend))
+        export_line = f'export PATH={quoted_prepend}:"$PATH"'
+        home = Path(tempfile.mkdtemp(prefix="coder-eval-codex-home-"))
+        for name in _LOGIN_PROFILE_NAMES:
+            content = self._login_profile_content(name, original_home, export_line)
+            (home / name).write_text(content, encoding="utf-8")
+        self._login_shell_home = home
+        self._log.debug(f"Login-shell mock-PATH home: {home}")
+
+    @staticmethod
+    def _login_profile_content(profile_name: str, original_home: str, export_line: str) -> str:
+        """One generated profile file: source the ORIGINAL home's profile (so
+        image/user setup isn't lost), then re-prepend the mock dirs.
+
+        ``.bash_profile`` mimics bash's first-found chain over the original
+        home; the ``.profile`` twin (read by ``sh``/``dash`` login shells)
+        sources only ``.profile`` — the bash-specific files may contain
+        bashisms a POSIX shell would choke on.
+        """
+        lines = [
+            "# Generated by coder_eval (CodexAgent): /etc/profile resets PATH in",
+            "# login shells, dropping the mock-CLI prepend — this restores it.",
+        ]
+        if original_home:
+            orig = shlex.quote(original_home)
+            if profile_name == ".bash_profile":
+                lines += [
+                    f"if [ -r {orig}/.bash_profile ]; then . {orig}/.bash_profile",
+                    f"elif [ -r {orig}/.bash_login ]; then . {orig}/.bash_login",
+                    f"elif [ -r {orig}/.profile ]; then . {orig}/.profile",
+                    "fi",
+                ]
+            else:
+                lines += [
+                    f"if [ -r {orig}/.profile ]; then . {orig}/.profile",
+                    "fi",
+                ]
+        lines.append(export_line)
+        return "\n".join(lines) + "\n"
+
+    def _cleanup_login_shell_home(self) -> None:
+        """Remove the generated per-task HOME (best-effort, idempotent)."""
+        home, self._login_shell_home = self._login_shell_home, None
+        if home is None:
+            return
+        shutil.rmtree(home, ignore_errors=True)
 
     def _build_thread_options(self) -> dict[str, Any]:
         """Build thread_start options from agent config.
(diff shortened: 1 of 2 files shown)


uipreliga (commented):
# Review: coder_eval — pr:26

Scope: pr:26 · branch `fix/codex-harness-shim-exec` · `b802957` · 2026-07-16T19:04Z · workflow variant

**Change class:** complex — introduces login-shell HOME shimming (temp-dir lifecycle, generated shell-profile files, PATH re-prepend) whose correctness depends on shell sourcing order, $HOME semantics, and shlex quoting

coder_eval is in exceptional health — a perfect 10/10 with clean type safety, tests, security, architecture, and harness scoring, and the only blemishes are two low-severity maintainability nits in the Codex agent that pose no correctness, security, or scoring risk, so the bottom line is ship-as-is while tidying the two codex_agent.py items opportunistically.

## Summary

| 1. Code Quality & Style | 9.9 / 10 | 0 | 0 | 0 | 1 | Implicit magic-string coupling between _LOGIN_PROFILE_NAMES tuple and the ".bash_profile" literal branch |
| 7. API Surface & Maintainability | 9.9 / 10 | 0 | 0 | 0 | 1 | _build_codex_env docstring omits the HOME/CODEX_HOME override it now performs |

**Overall Score**: 10 / 10 · **Weakest Axis**: Code Quality & Style at 9.9 / 10

## Blockers

None.

None.

## Nits

1. **[Axis 1] Implicit magic-string coupling between _LOGIN_PROFILE_NAMES tuple and the ".bash_profile" literal branch** (`src/coder_eval/agents/codex_agent.py:1168`) — The set of generated profile files is declared once as `_LOGIN_PROFILE_NAMES = (".bash_profile", ".profile")` (line 223) and driven through the loop `for name in _LOGIN_PROFILE_NAMES:` (line 1135), but `_login_profile_content` distinguishes the two kinds by re-matching the literal string: `if profile_name == ".bash_profile":` (line 1168), with everything else falling into the sh/.profile `else`. This is a second source of truth: adding e.g. `.zprofile` to the tuple would silently route it through the POSIX `.profile` branch (which sources only `.profile`, not the bash first-found chain), with no error. Minor for two files, but the bash-vs-POSIX distinction would be clearer as an explicit flag/parameter (e.g. pass `is_bash=name == ".bash_profile"` at the call site, or key a small mapping) rather than re-deriving it from the filename inside the content builder.
2. **[Axis 7] _build_codex_env docstring omits the HOME/CODEX_HOME override it now performs** (`src/coder_eval/agents/codex_agent.py:1063`) — The method-level docstring at lines 1063-1074 still enumerates only the two env vars the method used to set: it says the env "Carries the API key (``CODEX_API_KEY``...) and, when the sandbox resolved ``mock_path_dirs``, a PATH with those directories prepended..." — but the method now also injects `env["HOME"] = str(self._login_shell_home)` and `env["CODEX_HOME"] = str(self._codex_home())` (lines 1090-1091). Overriding HOME is the single most surprising side effect in this diff: it applies to the whole Codex app-server process (not just the `bash -lc` shells the design targets), which is a maintainability/discoverability trap for a future reader who reads the docstring as the method's contract. The behavior IS explained by the inline comment at lines 1086-1089, but the docstring — the first thing a maintainer reads — is now stale/incomplete. Add one sentence to the docstring noting that HOME is repointed at the per-task login-shell profile dir (with CODEX_HOME pinned back to the real codex state dir so auth/rollouts are unaffected), cross-referencing `_setup_login_shell_home`.

## What's Missing

**Parallel paths:**
- 🟡 The login-shell PATH-restore guard lives only in CodexAgent, but ClaudeCodeAgent (claude_code_agent.py:706,1062) and antigravity_agent (antigravity_agent.py:314,378-396) consume the SAME `env_path_prepend` mock-PATH prepend the sandbox hands every agent. If either spawns a login shell (`bash -lc` or any `-l`), the identical `/etc/profile` PATH reset drops the mock prepend and bare commands resolve to the REAL CLIs (real-tenant contamination) with no error. The PR should state that only Codex issues login shells, or the guard belongs on the shared Agent base rather than one agent. _(trigger: src/coder_eval/agents/codex_agent.py)_

**Daily/nightly:**
- 🟡 The fix exists for the containerized nightly Codex path (mock CLIs shadowing real-tenant CLIs) and hard-codes container-shell assumptions: that `/etc/profile` resets PATH and that `~/.bash_profile` is sourced AFTER it. If the nightly base image's `/etc/profile` or default login shell differs, the shim silently no-ops and contamination returns with no failure signal. The PR does not state whether the nightly container image must be rebuilt or validated against these assumptions before the fix is relied on. _(trigger: src/coder_eval/agents/codex_agent.py)_
- 🔵 `_build_codex_env` repoints HOME for the ENTIRE codex app-server process in the nightly container, not just the `bash -lc` shells the design targets; any `$HOME`-relative behavior of the codex binary itself now resolves to the ephemeral temp dir (only CODEX_HOME is pinned back). The whole-process blast radius of this override is undocumented and shares root cause with the Axis 7 docstring-omission finding. _(trigger: src/coder_eval/agents/codex_agent.py)_ _(restates: Axis 7: _build_codex_env docstring omits the HOME/CODEX_HOME override)_

**Tests:**
- 🔵 The only real-shell end-to-end test (`test_login_shell_restores_mock_prepend_end_to_end`) exercises `bash -lc` exclusively; the generated `.profile` twin (read by `sh`/`dash` login shells) is asserted at file-content level only and never executed by a real `sh`/`dash` login shell, so a sourcing bug or accidental bashism specific to the sh path would pass unit tests. _(trigger: tests/test_codex_agent.py)_

- (none)

- Add a parametrized unit test over `_LOGIN_PROFILE_NAMES` in `tests/` that, for every entry in the tuple, invokes `_login_profile_content(name)` and asserts the generated content matches that name's expected shell semantics — the bash first-found chain (`.bash_profile`/`.bashrc`) vs. the POSIX `.profile` sourcing — rather than asserting only that content is non-empty. This makes the tuple↔content-builder coupling self-checking: adding e.g. `.zprofile` to `_LOGIN_PROFILE_NAMES` without teaching `_login_profile_content` about it (so it silently falls into the POSIX `else` branch) trips a failing assertion instead of shipping wrong login-shell wiring. Ideally the test also asserts `_login_profile_content` has an explicit branch/mapping entry for each name (no unhandled name falls through to a generic default). _Why not static:_ The correct content for a given profile filename is a semantic property of shell login-file sourcing behavior (which chain a shell reads, in what order), not a grep-/AST-detectable pattern. A lint rule can't judge whether the `else` branch is right for a newly added filename; only executing the content builder and checking the emitted script against expected sourcing semantics can. And the underlying 'literal re-derived instead of parameterized' shape is exactly the kind of second-source-of-truth design judgment a mechanical rule would either miss or over-flag. _Prevents:_ Finding 1 (implicit magic-string coupling between `_LOGIN_PROFILE_NAMES` and the `.bash_profile` literal branch in `_login_profile_content`) — the silent mis-routing of a future profile name through the wrong shell branch.

## Top 5 Priority Actions

1. None of the confirmed findings can change a task's score or final_status for identical agent output, so there are no scoring-integrity fixes to front-load — proceed in pure impact order below.
2. Remove the second source of truth in src/coder_eval/agents/codex_agent.py:1168 by passing an explicit `is_bash` flag (or keying a small mapping) from the `_LOGIN_PROFILE_NAMES` loop into `_login_profile_content`, so adding a profile like `.zprofile` can't silently fall through the POSIX `.profile` branch.
3. Update the `_build_codex_env` docstring at src/coder_eval/agents/codex_agent.py:1063 to document the new HOME/CODEX_HOME repointing side effect (HOME → per-task login-shell profile dir, CODEX_HOME pinned to the real state dir), cross-referencing `_setup_login_shell_home`, since it is the most surprising behavior in the diff and the docstring is a maintainer's first contract.
4. Convert the magic-string coupling into permanent enforcement by adding a custom lint rule (tests/lint/rules/, CE-series) that flags re-deriving branch behavior from a literal already enumerated in a canonical tuple, per CLAUDE.md's 'could a lint rule have prevented this?' guidance.
5. Hold the line on the perfect axes by keeping the merge-unification, reconciliation-invariant, and classification-aggregate test guards green as the Codex/login-shell surface evolves, treating any new drift between docstrings and behavior as a regression.

---

bai-uipath (approved):
**Approve** — correct, well-contained, and now validated end-to-end in the real nightly container. Notes below are optional.

## What we verified

- **Unit + CI:** all 15 login-shell tests pass (including the real-`bash -lc` e2e that simulates the `/etc/profile` reset), and CI's "Live Integration Tests (Codex)" is green on Linux.
- **Real-environment e2e:** ran the actual PR-branch `CodexAgent` inside `skills-image:latest` — the container the nightly/weekly codex runs actually use — against its real `/etc/profile`, with and without the fix.

## Findings

- ✅ **The bug reproduces in-container.** In `skills-image`, a login shell resets PATH to the literal `/usr/local/sbin:…:/bin` and drops the mock-CLI prepend, so a bare `uip` resolves to the baked-in real `/usr/bin/uip`. That is exactly the real-tenant contamination the PR targets, confirmed in the true environment (macOS couldn't reproduce it — zsh + non-destructive `path_helper`; the VM host couldn't either — its profile augments rather than resets PATH).
- ✅ **The fix restores the prepend.** With `_setup_login_shell_home` + `_build_codex_env` composed the way the SDK merges the env over `os.environ`, a bare `uip` resolves to the mock and `$HOME` is correctly restored to `/root`. Together with the unit test that asserts `start()` folds HOME/CODEX_HOME/PATH into the `CodexConfig` env, the full chain is covered.
- ✅ **Blast radius is contained.** The HOME/CODEX_HOME redirect only activates when `_env_path_prepend` is set (mocked tasks) on POSIX — non-mocked codex tasks and the claude/antigravity agents are untouched (`codex_agent.py:1085`).
- ✅ **`CODEX_HOME` pin is right.** Redirecting HOME would otherwise move codex auth/rollout state into the temp dir; pinning it to `_codex_home()` (resolved in the parent's real env) keeps auth and sub-agent rollout recovery working.

src/coder_eval/agents/codex_agent.py
```diff
@@ -215,6 +217,11 @@ def _message_uncached_input(m: AssistantMessage) -> int:
 # supported"), so this is a fixed constant, not an operator knob.
 _CODEX_WIRE_API = "responses"
 
+# Login-shell profile files generated into the per-task HOME (see
+# _setup_login_shell_home). ``.bash_profile`` is what ``bash -l`` reads;
+# ``.profile`` covers ``sh``/``dash`` login shells.
+_LOGIN_PROFILE_NAMES = (".bash_profile", ".profile")
```
uipreliga: could we also make it work for Mac? (e.g. .zprofile)? how would it work for windows?

uipreliga (dismissed):

src/coder_eval/agents/codex_agent.py
dmorosanu: Added in 042976e.

**macOS:** codex picks the user's default shell - zsh on macOS - and runs commands as `zsh -lc`; `/etc/zprofile`'s `path_helper` resets PATH the same way `/etc/profile` does on Linux. The generated home now also ships `.zshenv`/`.zprofile`/`.zshrc`, and the app-server env points `ZDOTDIR` at it (zsh selects its dotfiles via `ZDOTDIR`, not `HOME`, so this survives the HOME-restore trick without ordering fragility). `.zshrc` is included because codex's shell snapshot sources it explicitly. Since `ZDOTDIR` stays exported, nested zsh login shells keep the prepend - the documented nested-shell gap is now bash/sh-only. Validated end-to-end with real zsh (login, interactive/snapshot shape, and nested invocations) against original dotfiles that all reset PATH.

**Windows:** nothing to generate - codex there never shells through bash; it uses `powershell -NoProfile -Command` (profiles skipped) or `cmd /c` (no profiles at all), so no profile chain ever resets PATH and the plain app-server env prepend survives as-is. The POSIX-only gate stays and the docstring now explains why.

PR #38, opened 2026-07-22
fix(reports): exact Student-t p-values and a paired comparison section

uipreliga:
## Why

`welch_t_test` treated the Welch t-statistic as a z-score (`2.0 * NormalDist().cdf(-t)`). The normal distribution has *lighter* tails than Student's t, so this **understated p** and manufactured significance exactly in the small-replicate regime coder_eval runs in. The in-code comment claimed the approximation "overestimates p slightly" — backwards. Measured: t=2.5 at df=4 is p=0.0668 (not significant); the old code reported 0.0124 (falsely significant). The 2-variant report fixture printed `p=0.005` where the exact value is 0.106 — understated ~20x.

## What changed

**Exact t-test.** Two-tailed Student-t p-value via the regularized incomplete beta (Lentz continued fraction, Numerical Recipes 6.4) plus proper Welch–Satterthwaite degrees of freedom. Pure stdlib — `math.lgamma`/`log1p` — so the module stays zero-dependency; no scipy. Cross-validated against `scipy.stats.ttest_ind(equal_var=False)` over 200 randomized group pairs, worst absolute difference 1.4e-13. Also fixes the docstring's claim of an exact fallback that never existed, and makes zero-variance-with-different-means return p=0.0 (a deterministic difference) instead of 1.0.

**Paired comparison.** Both variants run the same tasks, so an unpaired Welch test carries between-task difficulty variance in its denominator and is systematically underpowered. Adds `paired_t_test` and a `## Paired Comparison` section for 2-variant experiments. The existing paired block moves out of `## Replicate Statistics` and loses its `replicate_count > 1` gate, so single-replicate A/B — the common case — gets it too.

Pairing is over **per-task mean scores**. Replicate slots within a task share the task effect and are not independent; pairing them individually would understate the standard error. The task is the unit of analysis.

**One source of truth.** `_collect_variant_series` existed twice and had drifted: the markdown copy divides `duration_seconds` by `replicate_count` (correct — `VariantResult` sums it across replicates) while the HTML copy used it raw, so **the HTML report showed replicate-inflated durations**. Both now call one `collect_variant_series`. Likewise both reporters render one shared `paired_comparison()`, and the HTML report gains the paired section it never had.

**Coherent interval.** The paired section printed a percentile-bootstrap CI beside an exact-t p-value — two inference models on the same quantity, free to disagree, and degenerate over the 2–3 task means it resamples. `paired_t_ci` now derives the interval from the same distribution as the p-value, so the two always agree about whether 0 is excluded.

## Behavioural deltas (all intended)

- p-values in 2-variant reports get larger (honest) at small n; identical groups still return exactly 1.0.
- The 2-variant snapshot CI widens from `[-0.450, +0.200]` to `[-4.255, +4.005]` beside `p = 0.766`. The bootstrap was resampling two task means and reporting a falsely narrow interval; with 2 tasks you know almost nothing, and the interval now says so.
- A 2-variant experiment with only one common task explains why it has no paired result instead of rendering nothing.
- HTML durations are no longer replicate-inflated.

## Notes for review

- The `## Paired Comparison` note is deliberately **descriptive** and makes no power claim. Pairing trades degrees of freedom for correlation and is not unconditionally sharper — the committed 2-variant snapshot is itself a counter-example (Welch p=0.658, paired p=0.766).
- Tasks whose two variants ran unequal replicate counts are no longer excluded. That constraint only existed to zip replicate slots; under mean-pairing it discarded real data and biased toward fully-completed tasks.
- Non-finite guards: `student_t_two_tailed_p` fails closed to 1.0 (garbage must never read as significant), the list helpers return `None` (rendered `—`), and `regularized_incomplete_beta` raises rather than returning NaN.
- `tests/test_reports_stats_nonfinite.py` *enumerates* the module instead of listing helpers by hand, so a statistic added later is covered automatically. It caught the `regularized_incomplete_beta` NaN case while being written.
- Multiple-comparison correction is deliberately out of scope: both reporters gate p-values on exactly 2 variants, so there is no all-pairs family to correct.

## Testing

`make verify` — 3462 passed, 2 skipped, coverage 90.88%. Reference values pinned against scipy without taking a scipy dependency; both markdown snapshots regenerated and reviewed rather than loosened.


akshaylive (approved):
## Review: exact Student-t p-values and paired comparison section

### Overview
Replaces a normal-distribution approximation for the Welch t-test p-value with an **exact** Student-t p-value (via a from-scratch regularized incomplete beta function, stdlib-only), fixes a genuine correctness bug (the old code understated p-values, and its own in-code comment describing the error was backwards), adds a proper **paired comparison** section (per-task mean scores, exact paired t-test + t-based CI, replacing a percentile-bootstrap CI that could disagree with the p-value), and de-duplicates a series-collection helper that had drifted between the markdown and HTML reporters (fixing a real replicate-duration-inflation bug in the HTML report).

### Correctness (math)
Traced the incomplete-beta/continued-fraction implementation (`_betacf`, `regularized_incomplete_beta`, `student_t_two_tailed_p`) against the standard Numerical Recipes `betai`/`betacf` algorithm — faithful implementation, and the domain-boundary handling (`x<=0`→0, `x>=1`→1, `df<=0`→1.0 fail-closed) is correct. Hand-verified two of the new reference tests (Welch df=8/t=1.0/p≈0.347, and the 2-variant snapshot's duration row t≈2.83/df=2/p≈0.106) and both check out. The Welch–Satterthwaite df formula is textbook-correct. scipy cross-validation to 1.4e-13 (per PR description) is a strong signal.

### Strengths
- `tests/test_reports_stats_nonfinite.py` enumerates the module by introspection rather than hand-listing helpers — will catch a NaN-laundering regression in a future helper automatically.
- Fixing the HTML/markdown duration-divergence bug via one shared `collect_variant_series`/`VariantSeries` is the right fix (single source of truth), with a dedicated regression test (`test_html_and_markdown_report_the_same_duration`).
- The CI now shares its distribution with the p-value (`paired_t_ci`/`paired_t_test` both from `student_t_critical`/`student_t_two_tailed_p`), closing the "two inference models disagree" gap, with an explicit invariant test (`test_paired_ci_agrees_with_p_value`).
- Dropping the `replicate_count` equality gate on paired tasks (pairing per-task **means** instead of zipping replicate slots) is statistically correct and clearly documented.
- Snapshot deltas are reviewed/regenerated rather than loosened, and the PR body is unusually explicit about every behavioral delta.

### Minor, non-blocking notes
1. `student_t_critical`'s bracket-doubling search caps at `hi > 1e12` and returns that value rather than raising/logging. Unreachable at the `confidence=0.95` used everywhere today, but if a much higher confidence is ever passed, it would silently degrade to a wrong-width interval instead of failing loud, which is inconsistent with the "refuse rather than lie" philosophy used elsewhere in this PR (e.g. `bootstrap_mean_ci`'s explicit `ValueError`s). Worth a follow-up if `confidence` ever becomes caller-configurable.
2. `paired_comparison` silently drops tasks where either variant's per-task score list is empty, without surfacing that in the rendered section (the old code had an explicit "N task(s) excluded" message for its analogous exclusion). Probably rare in practice; a one-line note if it can happen would close the gap.

### Verdict
No blocking issues. Rigorous, well-tested statistics correctness fix — approving.

uipreliga:
Thanks for the thorough review, @akshaylive — both non-blocking notes were fair, so I addressed them rather than deferring:

**Note 1 — `student_t_critical` silently degrading past the bracket cap.**
Added a `logger.warning` at the `hi > 1e12` boundary so an extreme confidence fails loud instead of quietly returning a wrong-width interval, matching the "refuse rather than lie" stance used elsewhere in this PR. Still unreachable at the `confidence=0.95` used everywhere today, so no behavior change for real inputs — this just closes the gap if `confidence` ever becomes caller-configurable.

**Note 2 — `paired_comparison` silently dropping unpairable tasks.**
Added an `excluded_count` field to `PairedComparison` (= `|union of tasks| − |common paired tasks|`), and both the markdown and HTML reporters now append `(N task(s) excluded — not scored by both variants)` to the paired note when it's non-zero. When nothing is excluded the note is unchanged, so existing snapshots are untouched. Locked in with an assertion in `test_paired_comparison_ignores_tasks_with_no_scores`.

`make verify` is green (3462 passed, 2 skipped; typecheck + custom lint clean).

PR #62, opened 2026-07-28
chore(pages): replace the gh-pages docs mirror with a branded link stub

uipreliga:
## Why

`uipath.github.io/coder_eval/` was serving a **second live copy of `docs/`**, duplicating coder-eval.com/docs. Verified before this change:

- Both copies **self-canonicalized** — the mirror declared `<link rel="canonical" href="https://uipath.github.io/coder_eval/">` rather than deferring to coder-eval.com. Google elects one canonical per duplicate cluster, and nothing told it which to pick.
- `uipath.github.io/robots.txt` is **404**, so nothing blocked crawling, and the mirror shipped its own **`sitemap.xml` actively advertising all 14 URLs** for indexing.
- The mirror had **drifted**: it still served `/BYOD/` and `/CODEX_AGENT_GUIDE/`, pages renamed upstream to `DATASETS.md` and `agents/CODEX.md`. So the risk wasn't only duplication — the copy competing with us was out of date.

There's no penalty for duplicate content, but the mechanical costs are real: split link equity on a domain whose measured problem is a shortage of inbound links, and Search Console reporting "Duplicate, Google chose different canonical" for pages we're verified for — optimizing without feedback.

## What

Replaces the MkDocs deploy with a single branded stub page.

- **`.github/pages-stub/index.html`** — Coder Eval wordmark, a one-sentence description of what the product is, a link to the docs, and a brand-orange bar linking to this repo with its live star/fork counts.
- **`.github/workflows/docs.yml`** — publishes that stub instead of running `mkdocs gh-deploy`, force-replacing `gh-pages` so none of the old output (sitemap included) survives. `404.html` is a copy of the stub, so Pages serves it for every old deep URL without needing a file per route.

## Decisions worth reviewing

**`noindex, follow` rather than a canonical to `/docs`.** The stub's content differs from the target; Google discourages combining `noindex` with `rel=canonical` and ignores canonicals across substantially different pages. This takes the duplicate cluster to **zero** indexed pages rather than one.

**The mkdocs deploy is deleted, not commented out.** It had already been commented out once, and that is exactly how the duplicate got published — one uncomment away. `mkdocs.yml` itself is untouched and still required: CE028 generates the index surfaces from its `nav:`, the website derives its sidebar from it, and `mkdocs build` still verifies anchors locally. Only the *publishing* of a second site is gone.

**One logo copy, not two.** The wordmark is inlined from the website's `src/assets/logo-light.svg`. The light and dark variants differ in only four navy fills (three icon bars and the "Coder" text group — "Eval" and the arc are always brand orange), so those became `currentColor` and one 3.5 KB copy serves both colour schemes. The substitution was scripted with assertions that it recoloured exactly four fills and left `#FA4616` intact — no path data was hand-copied. The page also moved onto the real brand palette (`#FA4616`, `#10233A`, `#F5F7FA`) from my earlier approximations.

**Counts are never hardcoded.** A committed star count is wrong within days, so the markup ships spinners and the script fills them from `api.github.com`. Anything that isn't a live number stays a spinner — offline, non-2xx, missing field, or scripting disabled. Resolution is per cell, so a partial payload fills what it has. Accepted tradeoff: with JS off the spinner never resolves, which is the price of never publishing a stale number on a page nobody should land on. Unauthenticated API is 60 req/hour/IP.

An earlier revision carried a snapshot value plus restore-on-failure, a 4s `AbortController` deadline and a race guard; dropping the hardcoded values made all of it dead weight, so the script is now 10 lines.

**The page says what Coder Eval is.** It previously named the product and said the docs had moved, so anyone arriving from a stale link had to click through to learn what it was. It now leads with a one-sentence description, worded to match coder-eval.com and the package metadata so the product is described identically everywhere. There is a matching `<title>` and `meta description` — but note those are for link previews, **not ranking**: the page is `noindex`, so search engines will not surface it. A comment above them says so, so they aren't later mistaken for SEO copy doing work it isn't.

**No text is repeated.** The wordmark already renders "Coder Eval", so the `<h1>` beneath it was the same two words twice; the heading is now visually hidden, keeping document structure and serving as the accessible name (the logo is `aria-hidden`, which also stops a screen reader announcing the name twice). Linking `coder-eval.com/docs` inside the notice then made the "Documentation" list row a second copy of the same URL two lines apart, so that row is gone. Each destination appears exactly once: docs in the notice, the project site in the list, the repo in the bar.

## Verification

- Workflow YAML parses; the assemble step dry-run produces `index.html`, `404.html`, `.nojekyll`, `llms.txt`.
- All three outbound links return 200.
- Renders in both colour schemes (confirmed for light by deleting the dark `@media` rule at runtime, so the wordmark's `currentColor` was actually exercised).
- Counts: live load fills the real values; network reject, a request that never settles, and a payload missing `forks_count` each leave the spinner in place. Markup contains no numeric literals for the counts.
- `make lint` passes (166).

## After merge

This needs one **manual `workflow_dispatch` run** of the *Pages stub* workflow from the Actions tab — merging alone does not replace the live mirror.

## Not in scope

- The `sharp` HIGH Dependabot alert on `main` (transitive in `evalboard`, fixed in 0.35.0) is pre-existing and unrelated.
- `/llms.txt` on the website was pointing three links at the `uipath.github.io` mirror; that is fixed separately in the website repo, along with adding `/llms-full.txt`.

### DIFF
--- a/.github/pages-stub/index.html
+++ b/.github/pages-stub/index.html
@@ -96,14 +96,21 @@
         margin: 0 0 1rem;
         color: var(--muted);
       }
-      /* What the product is — the reason a visitor who lands here should care. */
+      /* What the product is — the reason a visitor who lands here should care.
+         With the h1 visually hidden this is the first thing read, so it carries
+         the visual weight a heading otherwise would. */
       .lead {
         color: var(--fg);
+        font-size: 1.0625rem;
       }
       /* Why this page exists; secondary to the description above it. */
       .notice {
         margin-bottom: 1.5rem;
         font-size: 0.9375rem;
+      }
+      .notice a {
+        color: var(--orange);
+        font-weight: 600;
       }
       ul {
         margin: 0 0 1.75rem;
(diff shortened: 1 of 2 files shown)


bai-uipath (approved):
**LGTM.** Deleting the mirror is right, and `404.html`-as-stub is the correct answer to 15 dead routes. All follow-up below.

- **Pull the wordmark out into a committed `.svg`.** 5.1 KB of inline SVG is 35% of the page, 3.1 KB of it font outlines for two words the hidden `<h1>` already renders. Fix: reference it root-absolute (`/coder_eval/logo.svg`), since `404.html` serves at arbitrary depths, and use a `<picture>` for light/dark because an external `` can't inherit `currentColor`; `website/src/assets/` already has both variants.

- **The multi-repo split is the real exposure, not the stub.** `docs/` is the source of truth here but publishing now lives entirely in `coder_eval_uipath/website/` via `scripts/sync-docs.mjs`. After this merges nothing in this repo publishes docs anywhere, and `mkdocs-material[imaging]>=9.5,<10` is pinned nowhere while `CLAUDE.md:208` still says to verify anchors with `mkdocs build`.

- **coder-eval.com is served from an unmerged branch.** `coder_eval_uipath` 62 has been open since Jul 22 at 73 commits and the live site tracks its head. The stub hardcodes `/llms-full.txt` and copies `logo-light.svg`, both of which exist only there. Not blocking (site is live, all links 200). Fix: land internal 62, and note the logo's source commit so it reads as a copy.

- **Merging doesn't take effect.** `workflow_dispatch`-only leaves the stale mirror live until someone remembers the Actions tab. Fix: `push:` on `main` with `paths: ['.github/pages-stub/**', '.github/workflows/docs.yml']`.

- **Minor:** `#FA4616` as text fails WCAG AA (3.53:1 on white, same for white-on-orange in the bar; rest of the palette is 5.5:1+); with JS off or rate-limited both counts spin forever, so consider dropping them; `noindex` is a no-op for the 15 deep paths, which now 404 on status.

PR #85, opened 2026-08-05
fix(docker): harness isolation via uid-drop barrier — close the criteria/grader leak

CarlesUIPath:
## Summary
Under `--driver docker`, the agent-under-test could read its own grading material —
`success_criteria`, the skills-repo graders (`check_*.py`), `RESOLUTION.md`, reference
agents, the staged `task.yaml`, and the per-task-dir mount. A colleague's nightly audit
measured ~2.4% of replicates reading the suite instead of solving the task (claude-code
highest); an adversarial "check /work" prompt drives it to ~100% across frontier and
open-weight models. Prior denylist mitigations (zip/base64/compile/delete) did not hold.

This installs an **allowlist permission barrier**: the container stays root, but the agent's
CLI subprocess is dropped to an unprivileged `agent` uid (2000). All grading material is
root-owned `chmod 0700` → **EACCES** to the agent, while grading (`SuccessChecker` /
`run_command` / judges) runs as root. The agent can read only its skill-docs and write its
own workspace.

## How it works
- **Agent-agnostic drop:** a `setpriv` shim; claude via `ClaudeAgentOptions.user`, codex via
  `launch_args_override`, antigravity via the `localharness` PATH-shadow. The orchestrator/
  grading process never leaves root.
- **Criteria strip + root-only channel:** the agent-readable `task.yaml` is criteria-stripped
  (`agent_safe_dump`); the real criteria/reference ride root-only in `task_full.json` and are
  merged back (raw-dict, before parsing) for grading.
- **Single choke point:** `container_perms.lock_harness_root_0700` / `grant_agent_ownership`,
  guarded by lint rule **CE033**.

| Leak surface | Closed by |
|---|---|
| Staged criteria (`/work/input`, `task_full.json`) | root-0700 lock + strip |
| Skills-repo graders (`check_*.py`, `RESOLUTION.md`, reference agents) | raw host mount forwarded + locked |
| Per-task-dir mount | root-0700 lock |
| Reference solution (`reference.file`/`directory`) | mount forwarded + locked |
| Signpost env (`TASK_DIR`, `SKILLS_REPO_PATH`) | live on the root grader env only |
| Baked image content | authoring invariant |

📖 **Setup & design reference:** the agent-uid ↔ root-grader arrangement, how tasks are
staged/authored under it, the barrier surfaces, the Docker-Desktop caveat, and the nightly
rollout are documented in `docs/DOCKER_ISOLATION.md` (extended in this PR).

## Verification

**New Docker test coverage (permanent, CI-gated).** Adds a root-in-container acceptance
suite (`tests/test_docker_user_separation.py`, run by the new `docker-isolation` CI job via
`make test-docker-isolation`): it forks a child dropped to the agent uid and asserts EACCES
on all six surfaces + a real dropped-CLI check (`id -u == 2000`, HOME writable, grader/
`task_full.json` unreadable) inside the freshly built image on **native Linux** — the
authoritative environment (see caveat).

**Independently confirmed the jailbreak reproduces → is closed.** Beyond the committed
suite, an out-of-tree reachability harness was run in two modes on the same image:
`reproduce` (barrier omitted) shows the six surfaces ARE readable to the dropped uid —
reproducing the jailbreak — and `verify-fixed` (barrier applied via the real primitives)
shows they become EACCES. The before/after pair proves the fix actually closes the leak
(and that the check isn't vacuous).

**Local:** gates green; the Docker acceptance suite passes; full `make test` = 4009 passed
with only pre-existing/environmental failures (live-API judge tests needing creds, local
`TASK_DIR`/task-tag pollution, a stale model-name pin, and report-command output strings
that live on `main`) — none touch this diff.

> ⚠️ **The barrier is Linux-authoritative.** On macOS/Windows Docker Desktop the
> gRPC-FUSE/VirtioFS uid-remap defeats `chmod 0700` on bind mounts, so local bind-mount
> checks are unreliable — trust the Linux CI job. (Locked mounts are `:rw` so the in-container
> chmod applies; the lock fails **loud** if it can't.)

## Out of scope / follow-ups
- `tempdir` driver + the Windows nightly slice (no container/uid boundary) — separate.

### DIFF
--- a/.claude/harness-candidates.md
+++ b/.claude/harness-candidates.md
@@ -107,3 +107,4 @@ Deferred lint/test guardrails surfaced during reviews. Promote to a `CExxx` rule
   caught them. The cleanup plan explicitly deferred this as YAGNI for the
   one-time purge, but any future doc rename/deletion re-opens the same blind
   spot — caught in the 2026-07-03 open-source-docs-cleanup implementation run.
+- [ ] CE033 form (c): a raw `.model_dump(` feeding the AGENT-READABLE `task.yaml` write in docker_runner must route through `agent_safe_dump` instead — no lint check today because distinguishing the stripped `task.yaml` write from the legitimate root-only `task_full.json` `model_dump` in the same function needs data-flow analysis, not a single-node AST match. Codebase currently compliant (task.yaml uses agent_safe_dump). — caught in docker-isolation-user-separation Phase 5.
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -137,7 +137,7 @@ jobs:
       - name: Run test suite with coverage
         run: |
           .venv/bin/pytest tests/ -v \
-            -m "not live and not lint" \
+            -m "not live and not lint and not docker_root" \
             --cov=coder_eval \
             --cov-report=term-missing \
             --cov-report=xml \
@@ -277,7 +277,7 @@ jobs:
         run: .venv/Scripts/pyright
 
       - name: Run test suite
-        run: .venv/Scripts/pytest tests/ -v -m "not live and not lint"
+        run: .venv/Scripts/pytest tests/ -v -m "not live and not lint and not docker_root"
 
       - name: Set up Node.js 20
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020  # v4.4.0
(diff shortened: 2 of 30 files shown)


CarlesUIPath:
Closing this PR — the uid-permission barrier isn't viable across OSs: on native Linux the chmod 0700 on the raw host bind mounts corrupts the real skills checkout (owner locked out, git status breaks), and on Docker Desktop the uid-remap defeats the lock so it doesn't protect at all. It's also SDK-version-coupled and fragile.

Another PR will follow up (copy/prune + grade-outside): the agent gets only a sanitized :ro bundle, the raw repo is never mounted, and grading runs outside the agent's reach.

PR #104, opened 2026-08-11
Isolate agent execution from grading data

dmorosanu:
## Summary

- Stage a sanitized agent task instead of passing the original task YAML into the agent container.
- Keep success criteria, references, lifecycle metadata, and the original task directory on the trusted host.
- Regrade the captured agent workspace after the container exits, using the original private task definition.
- Stop automatically mounting task and reference sources into the agent container.

## Why

The agent previously received the complete test case alongside the work it needed to perform. That allowed it to inspect private evaluation inputs such as criteria, solutions, references, and adjacent grading files. This change separates agent execution from grading while preserving the existing artifact-based result flow.

## Validation

- Ruff and whitespace checks passed.
- Podman probes confirmed a representative troubleshooting task does not expose task-local grading files, the original task directory, or private evaluation metadata to the agent container.

## Follow-up

The stacked public-input branch adds an explicit allowlist for plugin and template inputs, which closes the remaining exposure from raw source mounts.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -92,4 +92,7 @@ WORKDIR /work
 # /usr/local/bin/coder_eval_entrypoint.sh`, which makes the in-container
 # orchestrator launch robust to whatever a task image's Dockerfile declares.
 COPY docker/coder_eval_entrypoint.sh /usr/local/bin/coder_eval_entrypoint.sh
-RUN chmod +x /usr/local/bin/coder_eval_entrypoint.sh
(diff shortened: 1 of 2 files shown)


dmorosanu:
Parking this line of work: the container UID/GID isolation in #87 covers the target cases - grading material stays out of reach of the agent identity during the agent phase. #87 now runs agents without container-isolation support in normal mode and lets dynamic criteria run in the grader phase, so a separate harness-level data-separation layer is not needed right now. Branch stays parked for potential later reuse.

PR #105, opened 2026-08-11
Stage explicit public inputs for isolated agents

dmorosanu:
## Summary

- Add `sandbox.docker.agent_input_patterns` as an explicit public-file allowlist.
- Stage matching plugin and template files into a temporary public bundle instead of mounting their source directories.
- Rewrite agent-visible plugin and template paths to the staged bundle and leave original sources host-only.
- Document the end-to-end isolation architecture, lifecycle, and boundary between public inputs and private grading data.

## Why

Sanitizing the task alone was not enough: mounting whole plugin or template roots could still reveal evaluation files stored beside the agent-visible assets. This change gives each test an intentionally small, declared public input set.

## Validation

- Ruff and whitespace checks passed.
- Podman probes verified that allowlisted public files are available while adjacent private files are absent.
- Two unchanged skills tasks completed end-to-end in Podman using their original turn limits: Automation Discovery intake and UiPath Tasks negative guards.

## Stacked on

This draft is intentionally based on the agent/grader isolation branch so it can be reviewed as the second, narrower boundary-tightening step.


dmorosanu:
Parking with #104: superseded by the container UID/GID isolation in #87. Branch stays parked for potential later reuse.

PR #120, opened 2026-08-17
fix(deps): bump sqlparse 0.5.5 -> 0.6.0 to clear the pip-audit gate

mohsen-uipath:
Lockfile-only bump. `pip-audit` is a **Required** check and currently fails on every PR against this repo, `main` included — `sqlparse 0.5.5` accumulated four advisories after main's last green run, so the gate went red without a line of code changing.

Two are quadratic-blowup DoS vectors on the default parse path:

- **CVE-2026-71491** — `group_comments` is O(n²) on comment-only input, reachable via `format(sql, strip_comments=True)`.
- **CVE-2026-54284** — `TokenList.__init__` materializes `str(self)` per group, making nested grouping O(n·depth); a ~2 KB payload burns ~10 s of CPU.

Both are fixed in 0.6.0.

`sqlparse` is **transitive, not a direct dependency**: it arrives via `uipath-platform 0.1.8` under the optional `uipath` extra, whose constraint is `sqlparse>=0.5.5` — so 0.6.0 satisfies it and this is a pure `uv lock --upgrade-package sqlparse`, touching no other package.

## Verification

Ran CI's exact invocation against a `--extra dev --extra uipath --extra codex` sync:

```
$ .venv/bin/pip-audit --desc --skip-editable --ignore-vuln CVE-2026-4539 \
      --ignore-vuln CVE-2026-3219 --ignore-vuln PYSEC-2025-183
No known vulnerabilities found
```

`make verify` green on main plus this change: **4,162 tests**.

Found while triaging CI on #115, where this same gate was red for reasons unrelated to that PR.

### DIFF
--- a/src/coder_eval/orchestrator.py
+++ b/src/coder_eval/orchestrator.py
@@ -76,6 +76,28 @@ logger = logging.getLogger(__name__)
 _WAIT_FOR_GRACE_SECONDS = 2.0
 
 
+def _close_subprocess_transport(proc: asyncio.subprocess.Process | None) -> None:
+    """Release a finished subprocess's pipe transport deterministically.
+
+    ``asyncio`` gives ``Process`` no public way to do this: the transport is held
+    privately and is only closed when the object is garbage-collected. On
+    Windows' Proactor loop ``_ProactorBasePipeTransport.__del__`` then emits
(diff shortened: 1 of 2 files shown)


uipreliga (approved):
🚢

PR #126, opened 2026-08-18
test(early-stop): CE036 enforces the live_verdict determinism + monotonicity contract

mohsen-uipath:
Closes the two asks of #61 still open after #74/#78: **item 2's enforcement half** (the `live_verdict` contract was documented but nothing enforced it) and **item 4's design decision** (recorded in TASK_DEFINITION_GUIDE § `stop_early`: the ceiling/floor bounds deliberately stop at the armed subset — the weighted-score break is arm + weight + `stop_early_gate_threshold`).

**Zero runtime change**: the only `src/` diff is comments/docstrings in `criteria/base.py` (AST-verified identical to main with docstrings stripped). The deliverable is a CI-time gate: monotonicity over arbitrary Python is undecidable statically, so CE036 **replays** each live criterion against every prefix of recorded trajectories (authored ordering + seeded shuffles) and asserts determinism/monotonicity/no-raise directly, with registry-derived coverage (every `LiveSuccessCriterion` needs cases; every claimed polarity must be reached) so the fixture table can't decay into a vacuous always-`undecided` replay.

## Verification

**Sensitivity — mutation testing.** Six contract-breaking mutants injected into the real shipped checkers; CE036 caught all six with actionable messages, including a two-sided class (model polarity claim + checker drift together) that defeats the watcher's own polarity filtering: non-monotonic, non-deterministic (RNG), raising instead of degrading, polarity-dishonest, missing fixtures, vacuous all-undecided fixtures.

**Specificity + ecological validity.** Shipped checkers replay clean across all 15 fixtures and shuffles, and across **650+ (row × criterion) pairs extracted from real live activation-suite runs** (26 rows, one per skill stratum, real agent): 0 violations.

**Counterfactual harm — live probes (2nd commit).** Three NON-CI example tasks (`tasks/early_stop_contract_*.yaml`), each executed live twice — clean code vs a contract-breaking mutant. Every mutant **passed the entire pre-CE036 test surface** and silently flipped the live verdict; only CE036 fails:

| Probe | Clean code (live) | Mutant (live) | CE036 catch |
|---|---|---|---|
| `require_success` (crash must not count) | pass-stop on the real success → SUCCESS | false PASS on the crash, truncated at tool call 1 → FAILURE | `reaches` pin |
| `bounded_pass` (bounded pass not final) | fail-stop on the `max_count` overrun → FAILURE | false PASS froze a compliant count → **false SUCCESS** | NON-MONOTONIC + `reaches` |
| `any_engagement` (first-engagement regression) | pass-stop on the expected skill → SUCCESS | fail-stop on a foreign touch, truncated → FAILURE | NON-MONOTONIC |

The two trajectories not already pinned verbatim are added as `ContractCase`s, so the live-demonstrated walks stay enforced.

**Related context.** A kill-switch vs armed A/B on the activation suite (same 26 sampled rows, both arms) re-validated the runtime semantics this contract protects: 25/26 identical verdicts (the one diff traced to agent stochasticity — the watcher never fired), at −73% cost / −60% wall time. The tc-1 pass-stops that deliver those savings are trustworthy precisely because `live_verdict` is monotone — which is what CE036 keeps true as new live criteria are added.

**Gates**: `make verify` green — 4,165 tests, 358 lint (CE036 included), pyright 0 errors (incl. the new explicit pass over the contract engine).

**Honest limits** (documented on the rule): replay proves the contract on supplied trajectories, not in general; in-tree union types only — and since `tests/` isn't shipped in the wheel, docs/EXTENDING.md tells plugin authors to copy the replay pattern (a `ContractCase`-style fixture + the prefix walk) into their own test suite, with `tests/lint/live_verdict_contract.py` as the reference implementation.


bai-uipath (approved):
**Drop the three example tasks; fix the rest if you agree, otherwise lgtm.** The replay approach is right — monotonicity isn't statically decidable, and the registry-derived coverage checks are what keep the fixture table from rotting into a vacuous always-`undecided` pass. That part is clean.

**Drop the three `tasks/early_stop_contract_*.yaml`.** The mutants aren't checked in, so running them only shows the clean-code half — the counterfactual can't be reproduced from the repo, and the enforced version of that evidence is already the two new `ContractCase`s. One of them also fails by design in the directory people point `coder-eval run` at. Fix: delete all three; keep the narrative in the PR body.

**The second pyright invocation analyzes zero files.** Pyright's `exclude` beats an explicitly-passed CLI file arg, so the extra `make typecheck` pass exits 0 having checked nothing — I confirmed with a probe file (`filesAnalyzed: 0` under `tests/`, `1` under `src/`). Both new modules are clean when actually checked, so this costs nothing to fix. Fix: a second config file with those paths in `include`, or drop `"tests"` from the exclude list.

**Drop the shared telemetry-builder module.** It shares two six-line constructors between exactly two files, and the wrappers still have to reconcile two different `tool_id` conventions — while the judge test files keep their own builders. Fix: build the telemetry inline in the contract module and revert the early-stop test file entirely.

**Minor:** the empty-table test hardcodes today's two type names, so adding a third live criterion reds an unrelated test at exactly the wrong moment — compare against the registry instead. A raise on the final prefix leaves the terminal verdict stale and stacks a bogus `reaches` violation on the real one.

**Worth considering:** you extracted 650+ real activation-suite pairs to validate this and checked in none. Three of them as `ContractCase`s would move the standing gate off 38 hand-written tool calls onto trajectory shapes nobody would think to author.

mohsen-uipath:
Thanks — four fixed, one left for your call.

**Pyright.** Confirmed, and worse than it looked: `include` is a silent no-op too, CI never ran that pass at all (it calls `pyright` directly, not `make typecheck`), and once actually checking it caught six real `reportImplicitStringConcatenation` errors in both modules. Now a config generated from `[tool.pyright]` (`tests/lint/pyright_config.py`) so the passes cannot drift, wired into Linux + Windows CI. The PR body's "pyright 0 errors" claim was true only because that pass read nothing — I will correct it.

**Final-prefix raise.** Fixed. The existing raise test passed only by accident of its fixture, so the case had no coverage; it does now.

**Hardcoded type names.** Fixed — compared against the registry.

**The three probe tasks.** Dropped. The `ContractCase`s stay and now cite this PR for provenance.

**Shared telemetry module.** Left as is — your facts hold (2 of 20 files). The one thing worth keeping is the frozen timestamp, which CE036 needs. Say the word and the builders go inline in the contract module and `test_early_stop.py` reverts.

**Real trajectories.** Agreed, follow-up issue rather than holding this PR.

**One more:** `permuted_violations` kept the original `sequence_number`s after shuffling. The watcher sorts by that field, so the layer was probing unreachable inputs — and would go silently no-op for any checker that sorts by it. Now renumbered `0..N-1`, pinned by a test that returns nothing against the pre-fix engine.

`make verify` green: 4,162 tests, 357 lint, pyright 0 errors on both passes.

PR #127, opened 2026-08-18
fix(codex): store command output whole in result_summary

joeysbase:
## Problem

The Codex agent truncated command output with `f"Output: {output[:100]}"` when building `CommandTelemetry.result_summary`. Because `result_tokens` is derived from `result_summary` length (`ceil(len/4)`), this pinned ~77% of Bash results near a ~31-token floor and **under-reported every Codex tool result**, skewing the cost model.

This also violated the codebase-wide untruncated-`result_summary` contract (the Claude agent already stores full output; `test_resolve_pending_command_does_not_truncate_long_result_summary` asserts a 5000-char summary survives). Codex was the outlier.

## Change

- **`codex_agent.py`**: store command output whole in `result_summary` (display trimming belongs in renderers/reports, not capture).
- **`CE043` lint rule** (`tests/lint/rules/ce043_no_command_output_truncation.py`, wired into `runner.py`): forbids constant-slice truncation of captured command output (`output[:N]`, `.stdout[:N]`, …) inside `agents/`, so the contract can't silently regress. Scoped to `agents/`, with a `# noqa: CE043` escape hatch.
- Regression test (`test_command_output_recorded_whole_not_truncated`) + full lint-rule unit tests.

## Code-review fixes applied on top

- **`streaming/renderers.py`**: cap the Rich live-feed `ToolEnd` preview at `_MAX_RESULT_LEN` so a now-whole `result_summary` can't flood the console; the reported char count still shows the true full length (+ regression test).
- **`models/telemetry.py`**: narrow the `result_tokens` docstring — the untruncated contract covers captured command stdout/stderr, not the intentionally-brief one-line summaries of non-command tool items (`_summarize_tool_item`).
- **`experiments/default.yaml`**: `max_turns` raised to `100` (intentional, for larger task runs) with the comment clarified as a hard safety ceiling; `task_timeout`/`turn_timeout` remain the practical guards.

## Verification

`make verify` green — ruff, pyright (0 errors), custom lint (incl. CE043), 4210 tests, coverage 91.72%.


bai-uipath (approved):
**Approve.** Clean, well-scoped fix — restoring the untruncated contract is right, and CE043 plus the regression test keep it from drifting back. The renderer cap reporting the true char count while trimming the body is the correct split.

Minor, non-blocking: the log-file renderer still writes `result_summary` whole, so every Codex Bash result now lands in task.log at full length. Pre-existing for Claude, and arguably what a log should do — worth a deliberate call rather than an omission.

PR #133, opened 2026-08-21
fix(deps): bump anthropic to 1.0.0, migrate Bedrock judge path to httpx2

akshaylive:
## Summary
- Bump `anthropic` to `>=1.0.0,<2.0.0` (was `>=0.86.0`); anthropic 1.0.0 migrated its HTTP layer from `httpx` to `httpx2`, so its exceptions (e.g. `APIConnectionError`) now carry an `httpx2.Request`.
- Declare `httpx2>=2.12.0` explicitly in `pyproject.toml` since `judge_bedrock.py` constructs it directly (previously only pulled in transitively).
- Migrate `judge_bedrock.py`'s Bedrock invoker off the classic `httpx` package onto `httpx2` for consistency with the new anthropic SDK, updating `test_judge_bedrock.py`, `test_judge_anthropic.py`, and doc references accordingly.
- Relocked `uv.lock`.

## Test plan
- [x] `make test` (full suite): 4690 passed, 8 skipped (missing API keys/docker), 3 pre-existing environment-only failures unrelated to this change (Docker daemon unavailable; live Codex/Claude Code agent network tests)
- [x] `ruff check` + `pyright` clean on all touched files

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -76,7 +76,7 @@ jobs:
 
       - name: Install uv
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade "pip>=26.2"  # PYSEC-2026-3721 fix floor
           pip install uv
 
       - name: Install project dependencies (hash-verified from uv.lock)
@@ -297,7 +297,7 @@ jobs:
 
       - name: Install uv
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade "pip>=26.2"  # PYSEC-2026-3721 fix floor
           pip install uv
 
       - name: Install project without [uipath] extra
@@ -381,7 +381,7 @@ jobs:
 
       - name: Install uv
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade "pip>=26.2"  # PYSEC-2026-3721 fix floor
           pip install uv
 
       - name: Install project dependencies (hash-verified from uv.lock)
@@ -527,7 +527,7 @@ jobs:
 
       - name: Install uv
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade "pip>=26.2"  # PYSEC-2026-3721 fix floor
           pip install uv
 
       - name: Install project dependencies (hash-verified from uv.lock)
@@ -697,7 +697,7 @@ jobs:
 
       - name: Install uv
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade "pip>=26.2"  # PYSEC-2026-3721 fix floor
           pip install uv
 
       - name: Install project dependencies (hash-verified from uv.lock)
@@ -840,7 +840,7 @@ jobs:
 
       - name: Install uv
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade "pip>=26.2"  # PYSEC-2026-3721 fix floor
           pip install uv
 
       - name: Install project dependencies (with codex extra)
@@ -919,7 +919,7 @@ jobs:
 
       - name: Install uv
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade "pip>=26.2"  # PYSEC-2026-3721 fix floor
           pip install uv
 
       - name: Install project dependencies (hash-verified from uv.lock)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,11 +34,14 @@ dependencies = [
     "click>=8.3.3",
     "rich>=14.3.3",
     "python-dotenv>=1.2.2",
+    # Cap the major: 1.0.0 already proved a major can drop call kwargs
+    # (temperature/top_p/top_k off messages.create — see judge_anthropic.py).
     "anthropic>=1.0.0,<2.0.0",
     # anthropic>=1.0.0 migrated its HTTP layer from httpx to httpx2 (its own
     # exceptions, e.g. APIConnectionError, now carry an httpx2.Request). Declared
-    # explicitly since our code/tests construct httpx2 types directly.
-    "httpx2>=2.12.0",
+    # explicitly since our code/tests construct httpx2 types directly. Capped to
+    # mirror anthropic's own bound on it (`httpx2<3,>=2.0.0`).
+    "httpx2>=2.12.0,<3.0.0",
     "claude-agent-sdk>=0.2.124",
     "anyio>=4.13.0",
     "radon>=6.0.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,11 @@ dependencies = [
     "click>=8.3.3",
     "rich>=14.3.3",
     "python-dotenv>=1.2.2",
-    "anthropic>=0.86.0",
+    "anthropic>=1.0.0,<2.0.0",
+    # anthropic>=1.0.0 migrated its HTTP layer from httpx to httpx2 (its own
+    # exceptions, e.g. APIConnectionError, now carry an httpx2.Request). Declared
+    # explicitly since our code/tests construct httpx2 types directly.
+    "httpx2>=2.12.0",
     "claude-agent-sdk>=0.2.124",
     "anyio>=4.13.0",
     "radon>=6.0.1",
--- a/src/coder_eval/evaluation/judge_anthropic.py
+++ b/src/coder_eval/evaluation/judge_anthropic.py
@@ -41,6 +41,13 @@ async def invoke_anthropic_judge_async(
     caller can reuse ``extract_verdict_from_anthropic_response`` — Anthropic's
     native shape (content blocks with ``type: tool_use``) is identical
     between this SDK call and the Bedrock httpx2-direct call.
+
+    ``temperature`` is forwarded via ``extra_body`` rather than as a top-level
+    kwarg: anthropic 1.0.0 dropped ``temperature``/``top_p``/``top_k`` from
+    ``AsyncMessages.create``'s typed signature, but the Messages API itself
+    still accepts ``temperature`` in the raw JSON body — the same body shape
+    the Bedrock path already sends it in — so ``extra_body`` keeps both judge
+    backends honoring ``LLMJudgeCriterion.temperature`` identically.
     """
     alias = to_anthropic_alias(model)
     client = AsyncAnthropic(timeout=timeout_seconds)
@@ -51,12 +58,17 @@ async def invoke_anthropic_judge_async(
                 system=system,
                 messages=[{"role": "user", "content": user}],
                 max_tokens=max_tokens,
-                temperature=temperature,
-                tools=[tool_spec],  # type: ignore[arg-type]
+                tools=[tool_spec],  # pyright: ignore[reportArgumentType]
                 tool_choice={"type": "tool", "name": tool_spec["name"]},
+                extra_body={"temperature": temperature},
             )
         except APIError as e:
             # The SDK already retries transient failures internally (2 attempts
             # by default) — do not add another retry loop here.
             raise JudgeInfrastructureError(f"Anthropic judge API error: {e}") from e
+        except Exception as e:
+            # A signature/contract break (e.g. a removed or renamed kwarg after an
+            # SDK bump) must not be scored as an agent failure — see CLAUDE.md's
+            # CE039 rationale: an eval-infra fault is not the agent's fault.
+            raise JudgeInfrastructureError(f"Anthropic judge call failed: {e}") from e
     return response.model_dump()
--- a/src/coder_eval/evaluation/judge_anthropic.py
+++ b/src/coder_eval/evaluation/judge_anthropic.py
@@ -40,7 +40,7 @@ async def invoke_anthropic_judge_async(
     Returns the SDK response converted to a dict via ``model_dump`` so the
     caller can reuse ``extract_verdict_from_anthropic_response`` — Anthropic's
     native shape (content blocks with ``type: tool_use``) is identical
-    between this SDK call and the Bedrock httpx-direct call.
+    between this SDK call and the Bedrock httpx2-direct call.
     """
     alias = to_anthropic_alias(model)
     client = AsyncAnthropic(timeout=timeout_seconds)
--- a/src/coder_eval/evaluation/judge_bedrock.py
+++ b/src/coder_eval/evaluation/judge_bedrock.py
@@ -14,7 +14,7 @@ agent_judge uses the Claude Code SDK subprocess instead — the two paths
 intentionally do not share an HTTP client.
 
 Async on purpose: this is llm_judge's only implementation of the network
-call (there is no sync twin) — ``httpx.AsyncClient`` lets the call yield the
+call (there is no sync twin) — ``httpx2.AsyncClient`` lets the call yield the
 event loop instead of blocking a thread-pool thread for the wait, so
 ``SuccessChecker.check_all_async`` awaits it directly without pinning a
 thread. (``check_all_async`` currently runs criteria sequentially; running
@@ -27,7 +27,7 @@ import asyncio
 import logging
 from typing import Any
 
-import httpx
+import httpx2
 
 from coder_eval.errors import JudgeInfrastructureError
 from coder_eval.errors.categories import RetryConfig
@@ -90,13 +90,13 @@ async def invoke_bedrock_judge_async(
     attempts = _JUDGE_RETRY.max_retries + 1
     last_failure = ""
     last_exc: Exception | None = None
-    async with httpx.AsyncClient() as client:
+    async with httpx2.AsyncClient() as client:
         for attempt in range(attempts):
             if attempt:
                 await asyncio.sleep(compute_backoff(_JUDGE_RETRY, attempt - 1))
             try:
                 response = await client.post(url, headers=headers, json=body, timeout=timeout_seconds)
-            except httpx.HTTPError as e:
+            except httpx2.HTTPError as e:
                 last_failure = f"Bedrock invoke transport error: {e}"
                 last_exc = e
                 logger.warning("Bedrock judge attempt %d/%d failed: %s", attempt + 1, attempts, last_failure)
--- a/src/coder_eval/evaluation/verdict_tool.py
+++ b/src/coder_eval/evaluation/verdict_tool.py
@@ -40,7 +40,7 @@ SUBMIT_VERDICT_ANTHROPIC_TOOL: dict[str, Any] = {
     "description": _SUBMIT_VERDICT_DESCRIPTION,
     "input_schema": JudgeVerdict.model_json_schema(),
 }
-"""Anthropic-native tool spec for the Bedrock httpx-direct path and Anthropic SDK calls.
+"""Anthropic-native tool spec for the Bedrock httpx2-direct path and Anthropic SDK calls.
 
 Note: Anthropic uses ``input_schema``, not OpenAI's ``parameters``.
 """
@@ -132,7 +132,7 @@ def extract_verdict_from_anthropic_response(
     ``{"type": "tool_use", "name": "submit_verdict"}`` and validates the last
     one's ``input`` against ``JudgeVerdict``. Used by:
 
-    * The Bedrock httpx path (``invoke_bedrock_judge_async``) — raw JSON dict.
+    * The Bedrock httpx2 path (``invoke_bedrock_judge_async``) — raw JSON dict.
     * The Anthropic SDK Direct path (``invoke_anthropic_judge_async``) —
       response converted via ``Message.model_dump()``.
 
--- a/tests/test_judge_anthropic.py
+++ b/tests/test_judge_anthropic.py
@@ -10,6 +10,7 @@ from typing import Any
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
+from anthropic.resources.messages import AsyncMessages
 
 from coder_eval.evaluation.judge_anthropic import invoke_anthropic_judge_async
 from coder_eval.evaluation.verdict_tool import SUBMIT_VERDICT_ANTHROPIC_TOOL
@@ -29,10 +30,23 @@ def _make_response(*, score: float = 0.5, rationale: str = "ok") -> MagicMock:
 
 def _make_client(response: MagicMock | None = None) -> MagicMock:
     client = MagicMock()
-    client.messages.create = AsyncMock(return_value=response if response is not None else _make_response())
+    # spec-bound to the real ``AsyncMessages.create`` signature so a kwarg the
+    # installed SDK no longer accepts (e.g. a removed ``temperature``) fails
+    # here instead of silently passing against an unconstrained MagicMock.
+    client.messages = MagicMock(spec=AsyncMessages)
+    client.messages.create = AsyncMock(
+        wraps=lambda **kwargs: _bind_and_return(response if response is not None else _make_response(), **kwargs)
+    )
     client.__aenter__ = AsyncMock(return_value=client)
     client.__aexit__ = AsyncMock(return_value=None)
     return client
+
+
+def _bind_and_return(response: MagicMock, **kwargs: Any) -> MagicMock:
+    import inspect
+
+    inspect.signature(AsyncMessages.create).bind(MagicMock(), **kwargs)
+    return response
 
 
 async def _invoke(**overrides):
@@ -79,14 +93,33 @@ async def test_invoke_anthropic_judge_raises_on_empty_model() -> None:
 
 
 async def test_invoke_anthropic_judge_passes_temperature_and_max_tokens() -> None:
+    """``temperature`` travels via ``extra_body`` — anthropic 1.0.0 dropped it as a
+    top-level ``messages.create`` kwarg, but the Messages API still accepts it in
+    the raw JSON body (the same body shape the Bedrock path sends it in)."""
     client = _make_client()
     with patch("coder_eval.evaluation.judge_anthropic.AsyncAnthropic", return_value=client):
         await _invoke(temperature=0.7, max_tokens=321, system="sys", user="usr")
     kwargs: dict[str, Any] = client.messages.create.call_args.kwargs
-    assert kwargs["temperature"] == 0.7
+    assert "temperature" not in kwargs
+    assert kwargs["extra_body"] == {"temperature": 0.7}
     assert kwargs["max_tokens"] == 321
     assert kwargs["system"] == "sys"
     assert kwargs["messages"] == [{"role": "user", "content": "usr"}]
+
+
+async def test_invoke_anthropic_judge_escalates_on_signature_break() -> None:
+    """A kwarg the installed SDK no longer accepts must escalate as infra, not
+    silently score the row 0.0 (see judge_bedrock.py's parallel retry/escalation
+    contract and CLAUDE.md's CE039 rationale)."""
+    from coder_eval.errors import JudgeInfrastructureError
+
+    client = _make_client()
+    client.messages.create.side_effect = TypeError("create() got an unexpected keyword argument 'temperature'")
+    with (
+        patch("coder_eval.evaluation.judge_anthropic.AsyncAnthropic", return_value=client),
+        pytest.raises(JudgeInfrastructureError, match="Anthropic judge call failed"),
+    ):
+        await _invoke()
 
 
 async def test_invoke_anthropic_judge_wraps_api_error() -> None:
--- a/tests/test_judge_anthropic.py
+++ b/tests/test_judge_anthropic.py
@@ -90,13 +90,13 @@ async def test_invoke_anthropic_judge_passes_temperature_and_max_tokens() -> Non
 
 
 async def test_invoke_anthropic_judge_wraps_api_error() -> None:
-    import httpx
+    import httpx2
     from anthropic import APIConnectionError
 
     from coder_eval.errors import JudgeInfrastructureError
 
     client = _make_client()
-    sdk_error = APIConnectionError(request=httpx.Request("POST", "https://api.anthropic.com"))
+    sdk_error = APIConnectionError(request=httpx2.Request("POST", "https://api.anthropic.com"))
     client.messages.create.side_effect = sdk_error
     with (
         patch("coder_eval.evaluation.judge_anthropic.AsyncAnthropic", return_value=client),
--- a/tests/test_judge_bedrock.py
+++ b/tests/test_judge_bedrock.py
@@ -51,7 +51,7 @@ def _tool_use_response(score: float = 0.5, rationale: str = "ok") -> dict[str, A
 
 
 def _make_async_client(post_side_effect) -> MagicMock:
-    """Mock ``httpx.AsyncClient`` — supports the ``async with`` + repeated ``.post(...)`` shape."""
+    """Mock ``httpx2.AsyncClient`` — supports the ``async with`` + repeated ``.post(...)`` shape."""
     client = MagicMock()
     client.__aenter__ = AsyncMock(return_value=client)
     client.__aexit__ = AsyncMock(return_value=None)
@@ -83,7 +83,7 @@ async def test_invoke_bedrock_judge_happy_path(monkeypatch: pytest.MonkeyPatch) 
         captured["timeout"] = timeout
         return _make_response(status_code=200, json_data=_tool_use_response(score=0.5))
 
-    monkeypatch.setattr(judge_bedrock.httpx, "AsyncClient", lambda: _make_async_client(fake_post))
+    monkeypatch.setattr(judge_bedrock.httpx2, "AsyncClient", lambda: _make_async_client(fake_post))
     result = await _invoke(max_tokens=42)
 
     assert result["content"][0]["type"] == "tool_use"
@@ -99,7 +99,7 @@ async def test_invoke_bedrock_judge_happy_path(monkeypatch: pytest.MonkeyPatch) 
 
 async def test_invoke_bedrock_judge_raises_on_4xx(monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(
-        judge_bedrock.httpx,
+        judge_bedrock.httpx2,
         "AsyncClient",
         lambda: _make_async_client(lambda *a, **kw: _make_response(status_code=400, text='{"message":"bad model"}')),
     )
@@ -115,7 +115,7 @@ async def test_invoke_bedrock_judge_raises_on_5xx(monkeypatch: pytest.MonkeyPatc
         calls.append(1)
         return _make_response(status_code=500, text="upstream error")
 
-    monkeypatch.setattr(judge_bedrock.httpx, "AsyncClient", lambda: _make_async_client(counting_post))
+    monkeypatch.setattr(judge_bedrock.httpx2, "AsyncClient", lambda: _make_async_client(counting_post))
     with pytest.raises(JudgeInfrastructureError, match="Bedrock invoke failed: 500"):
         await _invoke()
     # Exactly 1 initial call + max_retries retries — read from the constant, don't hardcode.
@@ -125,7 +125,7 @@ async def test_invoke_bedrock_judge_raises_on_5xx(monkeypatch: pytest.MonkeyPatc
 
 async def test_invoke_bedrock_judge_raises_on_non_dict_response(monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(
-        judge_bedrock.httpx,
+        judge_bedrock.httpx2,
         "AsyncClient",
         lambda: _make_async_client(lambda *a, **kw: _make_response(json_data=["not a dict"])),
     )
@@ -149,12 +149,12 @@ async def test_invoke_bedrock_judge_raises_on_empty_model() -> None:
 async def test_invoke_bedrock_judge_wraps_transport_error(
     monkeypatch: pytest.MonkeyPatch, no_sleep: list[float]
 ) -> None:
-    import httpx as _httpx
+    import httpx2 as _httpx
 
     def raising_post(*a: Any, **kw: Any) -> MagicMock:
         raise _httpx.ConnectTimeout("connection timed out")
 
-    monkeypatch.setattr(judge_bedrock.httpx, "AsyncClient", lambda: _make_async_client(raising_post))
+    monkeypatch.setattr(judge_bedrock.httpx2, "AsyncClient", lambda: _make_async_client(raising_post))
     with pytest.raises(JudgeInfrastructureError, match="Bedrock invoke transport error") as excinfo:
         await _invoke()
     assert "connection timed out" in str(excinfo.value)
@@ -168,7 +168,7 @@ async def test_invoke_bedrock_judge_strips_v1_suffix_in_url(monkeypatch: pytest.
         captured["url"] = url
         return _make_response(json_data=_tool_use_response())
 
-    monkeypatch.setattr(judge_bedrock.httpx, "AsyncClient", lambda: _make_async_client(fake_post))
+    monkeypatch.setattr(judge_bedrock.httpx2, "AsyncClient", lambda: _make_async_client(fake_post))
     await _invoke(model="anthropic.claude-opus-4-6-v1")
     assert "/model/eu.anthropic.claude-opus-4-6/invoke" in captured["url"]
 
@@ -189,7 +189,7 @@ async def test_invoke_bedrock_judge_retries_429_then_succeeds(
         calls.append(1)
         return next(responses)
 
-    monkeypatch.setattr(judge_bedrock.httpx, "AsyncClient", lambda: _make_async_client(sequenced_post))
+    monkeypatch.setattr(judge_bedrock.httpx2, "AsyncClient", lambda: _make_async_client(sequenced_post))
     result = await _invoke()
     assert result["content"][0]["input"]["score"] == 0.9
     assert len(calls) == 3
@@ -205,7 +205,7 @@ async def test_invoke_bedrock_judge_403_fails_immediately(
         calls.append(1)
         return _make_response(status_code=403, text="forbidden")
 
-    monkeypatch.setattr(judge_bedrock.httpx, "AsyncClient", lambda: _make_async_client(counting_post))
+    monkeypatch.setattr(judge_bedrock.httpx2, "AsyncClient", lambda: _make_async_client(counting_post))
     with pytest.raises(JudgeInfrastructureError, match="Bedrock invoke failed: 403"):
         await _invoke()
     assert len(calls) == 1
@@ -215,7 +215,7 @@ async def test_invoke_bedrock_judge_403_fails_immediately(
 async def test_invoke_bedrock_judge_retries_connect_error_then_succeeds(
     monkeypatch: pytest.MonkeyPatch, no_sleep: list[float]
 ) -> None:
-    import httpx as _httpx
+    import httpx2 as _httpx
 
     calls: list[int] = []
 
@@ -225,7 +225,7 @@ async def test_invoke_bedrock_judge_retries_connect_error_then_succeeds(
             raise _httpx.ConnectError("connection refused")
         return _make_response(status_code=200, json_data=_tool_use_response())
 
-    monkeypatch.setattr(judge_bedrock.httpx, "AsyncClient", lambda: _make_async_client(flaky_post))
+    monkeypatch.setattr(judge_bedrock.httpx2, "AsyncClient", lambda: _make_async_client(flaky_post))
     result = await _invoke()
     assert result["content"][0]["type"] == "tool_use"
     assert len(calls) == 2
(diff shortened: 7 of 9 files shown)


uipreliga (commented):
# Review: coder_eval — pr:133

Scope: pr:133 · branch `akshaya/anthropic_upgrade` · `3fa04b5` · 2026-08-21T10:20Z · workflow variant

**Change class:** complex — a major-version SDK bump (anthropic 0.102.0 → 1.0.0) that swaps the SDK's HTTP layer (httpx → httpx2) and changes exception/type identities on the llm_judge network path; correctness requires reasoning about which SDK breaking changes were and were not handled.

Architecture, error handling, API surface and harness design are flawless (four axes at 10/10) and security is strong, but this PR's anthropic 1.0.0 bump ships a `temperature=` kwarg the new SDK rejects on the DEFAULT Direct judge path — turning a harness/SDK incompatibility into a silent `score=0.0` for every `llm_judge` row — and the three gates that should have caught it are all blind (a spec-less mock that positively asserts the dead kwarg, a blanket `# type: ignore` that hides the whole-call pyright error, and a real-SDK burn-in file no CI job selects), so with two PR-check jobs additionally red on the safe-chain min-age block for the 1-day-old `anthropic@1.0.0`, the bottom line is: do not merge until the judge call is fixed, gated, and the min-age window has aged out on its own rather than being waived.

## Summary

| 1. Code Quality & Style | 6.8 / 10 | 1 | 0 | 0 | 2 | anthropic 1.0.0 dropped `temperature` from `messages.create`, so the Direct-backend `llm_judge` path raises an uncaught TypeError and every judged row scores 0.0 |
| 2. Type Safety | 9 / 10 | 0 | 1 | 0 | 0 | Blanket `# type: ignore[arg-type]` on the `tools=` argument suppresses the whole `messages.create` call and is what kept `make typecheck` green over the removed `temperature` kwarg |
| 3. Test Health | 5.9 / 10 | 1 | 1 | 0 | 1 | Spec-less MagicMock Anthropic client lets the judge tests positively assert a call shape the real SDK rejects, so the suite is green over a fatal break |
| 4. Security | 9.4 / 10 | 0 | 0 | 1 | 1 | The org safe-chain minimum-package-age gate does not "either trip or land in the window" — it demonstrably TRIPS, and on exactly one distribution: `anthropic@1.0.0` (uv.lock:167, 1 day old). Two PR-check jobs are red because of it. The other three new distributions (`httpx2`/`httpcore2` at 3 days, `httpx2-jsfetch` at 14 days) pass the window. |

**Overall Score**: 8.9 / 10 · **Weakest Axis**: Test Health at 5.9 / 10
**Totals**: 🔴 2 · 🟠 2 · 🟡 1 · 🔵 4 across 8 axes.

## Blockers

1. **[Axis 1] anthropic 1.0.0 dropped `temperature` from `messages.create`, so the Direct-backend `llm_judge` path raises an uncaught TypeError and every judged row scores 0.0** (`src/coder_eval/evaluation/judge_anthropic.py:54`) — The PR treats the anthropic 0.102.0 -> 1.0.0 bump as a pure `httpx` -> `httpx2` rename, but the same major removed `temperature`/`top_p`/`top_k` from `AsyncMessages.create`. Line 54 still reads `temperature=temperature,` inside `await client.messages.create(`, and line 33 still declares `temperature: float,` on the public signature. Verified in the PR worktree: `inspect.signature(anthropic.resources.messages.AsyncMessages.create)` -> `['self','max_tokens','messages','model','cache_control','container','inference_geo','metadata','output_config','service_tier','stop_sequences','stream','system','thinking','tool_choice','tools','user_profile_id','extra_headers','extra_query','extra_body','timeout']` (no `temperature`, no `**kwargs`), and a real invocation raises `TypeError : AsyncMessages.create() got an unexpected keyword argument 'temperature'` before any network call. The Axis-1 defect is the misleading surface this leaves on a live path: `temperature` remains a documented parameter (line 33), is threaded in from `criteria/llm_judge.py:221` (`temperature=criterion.temperature`), is described on the user-facing model (`models/criteria.py:1344` "Sampling temperature for the judge model. 0.0 keeps grading deterministic."), and is asserted green by `tests/test_judge_anthropic.py:86` (`assert kwargs["temperature"] == 0.7`) — yet it can no longer reach the API at all. It also silently diverges the two parallel judge paths that CLAUDE.md treats as equivalent: `judge_bedrock.py:79` puts `"temperature": temperature` in the JSON body (still valid), while the Direct path now dies on it. Impact: `TypeError` is not in `criteria/base.py:81::_ESCALATING_EXCEPTIONS = (JudgeInfrastructureError, CheckerMisuseError)`, so `handle_criterion_errors_async` captures it into `CriterionResult(score=0.0, error=...)` — every `llm_judge` criterion on the Anthropic-Direct backend scores 0.0 instead of being graded or escalated. Fix: drop `temperature=temperature` from the `create(...)` call and either remove the parameter from `invoke_anthropic_judge_async` (and its call site at `llm_judge.py:221`) or route it through the supported `output_config`/`extra_body` surface; whichever is chosen, remove/retarget the now-false assertion at `tests/test_judge_anthropic.py:86` and build the test double as `MagicMock(spec=AsyncAnthropic)`-shaped so a removed kwarg fails the unit test instead of passing (`tests/test_judge_anthropic.py:31` currently uses a bare `MagicMock()`, which is why all 16 tests pass against a call that cannot execute).
2. **[Axis 2] Blanket `# type: ignore[arg-type]` on the `tools=` argument suppresses the whole `messages.create` call and is what kept `make typecheck` green over the removed `temperature` kwarg** (`src/coder_eval/evaluation/judge_anthropic.py:55`) — The call at lines 49-57 still passes a parameter `anthropic` 1.0.0 no longer accepts:

```
                max_tokens=max_tokens,
                temperature=temperature,            # line 54
                tools=[tool_spec],  # type: ignore[arg-type]   # line 55
```

Verified in the PR worktree: `inspect.signature(AsyncMessages.create)` on anthropic 1.0.0 yields `['self','max_tokens','messages','model','cache_control','container','inference_geo','metadata','output_config','service_tier','stop_sequences','stream','system','thinking','tool_choice','tools','user_profile_id','extra_headers','extra_query','extra_body','timeout']` — no `temperature`/`top_p`/`top_k` and no `**kwargs`. Calling `invoke_anthropic_judge_async(...)` in that worktree raises `TypeError: AsyncMessages.create() got an unexpected keyword argument 'temperature'`. `TypeError` is not in `criteria/base.py::_ESCALATING_EXCEPTIONS` (`(JudgeInfrastructureError, CheckerMisuseError)`), so `handle_criterion_errors_async` converts it into `CriterionResult(score=0.0)` — every `llm_judge` criterion on `DirectRoute` scores the agent 0.0 instead of escalating to `FinalStatus.ERROR`.

Why the type checker did not catch it: pyright ignores mypy's bracketed error codes and treats `# type: ignore[arg-type]` as a blanket line suppression, and it suppresses any diagnostic whose *range* covers that line. With `temperature` present the diagnostic is `reportCallIssue: No overloads for "create" match the provided arguments`, whose range spans the entire call (lines 49-57) — so the comment on line 55 swallows it. Reproduced with two identical probe functions differing only in that comment: without it, pyright reports the overload error; with it, `0 errors`. Replacing it with the rule-scoped, pyright-native form keeps the suppression narrow — with `# pyright: ignore[reportArgumentType]` on the `tools=` line, pyright still reports the `reportCallIssue` for `temperature`. Fix: (1) drop `temperature=temperature` from this call (anthropic 1.0.0 removed sampling params from the Messages surface entirely — `output_config.effort` is the replacement lever), and decide explicitly whether `LLMJudgeCriterion.temperature` becomes Bedrock-only (`judge_bedrock.py:79` still sends it in the request body) or is forwarded via `extra_body`; (2) change line 55 to `# pyright: ignore[reportArgumentType]` so this class of SDK break fails `make typecheck` instead of shipping; (3) consider `enableTypeIgnoreComments = false` in `[tool.pyright]` (or a `CEnnn` rule forbidding `# type: ignore` in favour of `# pyright: ignore[<rule>]`) — the repo has ~30 mypy-style `# type: ignore[code]` comments that are all silently blanket suppressions under the pyright-only setup.
3. **[Axis 3] Spec-less MagicMock Anthropic client lets the judge tests positively assert a call shape the real SDK rejects, so the suite is green over a fatal break** (`tests/test_judge_anthropic.py:86`) — `_make_client()` (line 30-35) returns a bare `client = MagicMock()` (line 31), so `client.messages.create(...)` accepts ANY keyword argument. Line 86 then positively certifies a kwarg the installed SDK rejects:

    85     kwargs: dict[str, Any] = client.messages.create.call_args.kwargs
    86     assert kwargs["temperature"] == 0.7

Verified in the prepared worktree at PR HEAD:
- `uv run python -c "import inspect; from anthropic.resources.messages import AsyncMessages; print(list(inspect.signature(AsyncMessages.create).parameters))"` on anthropic 1.0.0 → `['self','max_tokens','messages','model','cache_control','container','inference_geo','metadata','output_config','service_tier','stop_sequences','stream','system','thinking','tool_choice','tools','user_profile_id','extra_headers','extra_query','extra_body','timeout']` — no `temperature`, no `top_p`, no `**kwargs`.
- Calling the real function: `invoke_anthropic_judge_async(...)` → `TypeError: AsyncMessages.create() got an unexpected keyword argument 'temperature'`, raised before any network I/O.
- `uv run pytest tests/test_judge_anthropic.py tests/test_judge_bedrock.py -q -o addopts=""` → `16 passed in 0.44s`.

The offending production line is `src/coder_eval/evaluation/judge_anthropic.py:54` (`temperature=temperature,`), reached from `src/coder_eval/criteria/llm_judge.py:217-223` on the `DirectRoute` arm (the default backend). `TypeError` is not in `_ESCALATING_EXCEPTIONS` (`src/coder_eval/criteria/base.py:81` — only `JudgeInfrastructureError, CheckerMisuseError`), so `handle_criterion_errors_async` funnels it into `_failed_result` (`base.py:104`) which returns `score=0.0`. Every `llm_judge` criterion on the Direct backend therefore silently scores 0.0 while the run completes normally, and the test suite reports 100% line+branch coverage for `judge_anthropic.py`.

Fix: (a) stop mocking the client shape — build it with `MagicMock(spec=anthropic.resources.messages.AsyncMessages)` for `client.messages`, or add a signature-binding assertion (`inspect.signature(AsyncMessages.create).bind(client.messages, **create_kwargs)`) inside `_make_client`/the temperature test, so a removed or renamed SDK kwarg fails the unit test instead of passing it. (b) Delete the `temperature` assertion at line 86 once the production call drops the kwarg (anthropic 1.0.0 moves sampling params — `temperature` must go into `output_config`/`extra_body`, not a top-level kwarg), and add a test asserting the new carrier. Per the shared rubric item 17 ("Test mocks match real SDK shape") this whole file needs `spec=`-bound mocks. Mechanically detectable: propose a `CEnnn` lint rule forbidding a bare `MagicMock()` where the mock stands in for a third-party SDK client that production calls with keyword arguments (require `spec=`).
4. **[Axis 3] tests/test_judge_burn_in_live.py is selected by no CI step, so the anthropic 1.0.0 bump had no real-SDK gate at all** (`tests/test_judge_burn_in_live.py:50`) — Line 50 is `pytestmark = pytest.mark.live`, and the file docstring claims (lines 6-7) "Each test ``pytest.skip``s when the required credentials are not present, so the file is safe to run in CI without a credential set." It is never run in CI at all:
- `.github/workflows/pr-checks.yml:158` runs the suite with `-m "not live and not lint"` (and `:405` on Windows), excluding it.
- The `live-tests` job (`pr-checks.yml:645`) has `ANTHROPIC_API_KEY`, `AWS_BEARER_TOKEN_BEDROCK`, `AWS_REGION` in scope and a "tests actually ran (not silently skipped)" JUnit assertion, but its only two pytest steps (`:723`, `:735`) target `tests/test_claude_settings_enforcement_live.py`.
- `grep -rn "test_judge_burn_in_live" .github/ Makefile .claude/ docs/` returns zero hits; `grep -rn "burn_in|burn-in" --include="*.yml" --include="*.md" --include="Makefile" .` returns zero hits. Only `make test-live` (`Makefile:55`, `pytest -m live tests/`) reaches it, and `make verify` (`Makefile:69`) excludes live.

So `test_llm_judge_anthropic_direct_tool_channel` (line 73) — the one test that would have caught the `anthropic` 1.0.0 `temperature` removal confirmed in finding #1 — has never executed against this PR. This is a pre-existing wiring gap, but it is the direct reason a broken default-backend judge path shipped green, and the file is in scope for this PR.

Fix: add a step to the existing `live-tests` job (which already holds both credential sets and the JUnit "actually passed" parser) running `.venv/bin/pytest tests/test_judge_burn_in_live.py -m live -v --tb=short --strict-markers -ra --junit-xml=tmp/junit-burn-in.xml`, and extend the passed-count assertion to that XML so a silent credential-skip is a CI failure rather than a green run. Cost is 3 real LLM calls pinned to Sonnet 4.6 / Haiku 4.5 per the file's own docstring (lines 17-18).

1. **[Axis 4] The org safe-chain minimum-package-age gate does not "either trip or land in the window" — it demonstrably TRIPS, and on exactly one distribution: `anthropic@1.0.0` (uv.lock:167, 1 day old). Two PR-check jobs are red because of it. The other three new distributions (`httpx2`/`httpcore2` at 3 days, `httpx2-jsfetch` at 14 days) pass the window.** (`uv.lock:946`) — The lock adds three brand-new *distribution names* to the base (non-extra) runtime closure and bumps a fourth, all published within 3 days of this PR (today 2026-08-21): `uv.lock:946` `httpx2-2.12.0.tar.gz ... upload-time = "2026-08-18T13:22:09.086Z"`, `uv.lock:906` `httpcore2-2.12.0.tar.gz ... upload-time = "2026-08-18T13:22:08.211Z"`, `uv.lock:955` `httpx2_jsfetch-1.0.tar.gz ... upload-time = "2026-08-07T00:13:07.492Z"`, and `uv.lock:167` `anthropic-1.0.0.tar.gz ... upload-time = "2026-08-20T19:59:00.565Z"` — one day old. `truststore` also moves from the `uipath`-extra-only closure (it appears in `origin/main:uv.lock` only under `uipath`/`uipath-platform`) into the base closure via `uv.lock:904` `{ name = "truststore" }` under `httpcore2` and `uv.lock:944` under `httpx2`. This repo runs an org-mandated safe-chain minimum-package-age gate on every installing job (`.github/workflows/pr-checks.yml:35` `SAFE_CHAIN_MINIMUM_PACKAGE_AGE_EXCLUSIONS: ... || 'openai-codex-cli-bin,openai-codex'` and `docker/Dockerfile:68` `ARG SAFE_CHAIN_MINIMUM_PACKAGE_AGE_EXCLUSIONS="openai-codex-cli-bin,openai-codex,google-antigravity"`); none of the four new distributions is on either exclusion list, so the PR either trips that gate or lands inside its window with the risk unmitigated — and it records no decision either way. Mitigations that hold: every new entry is sha256-pinned in `uv.lock`; metadata confirms legitimate provenance (`httpx2`/`httpcore2` → `Project-URL: Homepage, https://github.com/pydantic/httpx2`, `Author-email: Tom Christie <tom@tomchristie.com>`, `License-Expression: BSD-3-Clause`), `anthropic 1.0.0` declares `Requires-Dist: httpx2<3,>=2.0.0` so the rename is upstream-driven, and `httpx2-jsfetch` is gated `marker = "sys_platform == 'emscripten'"` (verified not installed in the worktree venv). Fix: record in the PR body (a) that the safe-chain min-age gate was consciously waived or satisfied for `httpx2`, `httpcore2`, `httpx2-jsfetch`, `anthropic`, `idna`, and (b) the provenance check above; if the gate blocks, add the three names to the exclusion lists in the same PR rather than letting CI fail opaquely. Prefer holding the merge until PyPI attestations for `anthropic 1.0.0` / `httpx2 2.12.0` can be verified, since these land in the wheel every downstream `pip install coder-eval` consumer gets. CVSS:3.1/AV:N/AC:H/PR:H/UI:N/S:U/C:H/I:H/A:H

## Nits

1. **[Axis 1] Test-local `import httpx2 as _httpx` aliases the new library to the old library's name, in two duplicated function-local imports** (`tests/test_judge_bedrock.py:152`) — `tests/test_judge_bedrock.py:152` and `:218` both read `import httpx2 as _httpx`, then use `_httpx.ConnectTimeout` / `_httpx.ConnectError` / `isinstance(excinfo.value.__cause__, _httpx.ConnectTimeout)`. In a PR whose entire purpose is to move off `httpx`, aliasing `httpx2` to `_httpx` makes the grep for the old name look like a leftover and makes a reader believe the old client is still involved — the rest of the file (`:86`, `:102`, `:118`, `:128`, `:157`, `:171`, `:192`, `:208`, `:228`, `:240`) already spells it `judge_bedrock.httpx2`. Replace both with a single module-level `import httpx2` and use `httpx2.ConnectTimeout` / `httpx2.ConnectError` directly, dropping the two duplicated function-local imports.
2. **[Axis 1] Dependency block is internally inconsistent: new `httpx2` direct dependency ships with no major cap while the `anthropic<2.0.0` cap carries no rationale, against the block's own convention** (`pyproject.toml:37`) — Line 37 is `"anthropic>=1.0.0,<2.0.0",` and line 41 is `"httpx2>=2.12.0",`. Every other capped dep in this block states why it is capped — lines 49-54: `# Telemetry: ... Cap the major to guard against a 2.0 moving/removing either; the Azure exporter is a beta, so cap the minor since betas can break compatibility between releases.` above `"opentelemetry-sdk>=1.30.0,<2.0.0"` / `"azure-monitor-opentelemetry-exporter>=1.0.0b30,<1.1.0"`. Here the three-line comment at lines 38-40 explains the httpx2 addition but says nothing about the new `<2.0.0` cap on anthropic — and this PR is itself the proof that an anthropic major removes call kwargs, which is precisely the rationale worth recording. Conversely `httpx2` gets no cap even though the comment's own argument is a major-version coupling to anthropic, which pins it as `httpx2<3,>=2.0.0` (verified: `importlib.metadata.requires('anthropic')` includes `'httpx2<3,>=2.0.0'`). Make the two consistent: cap `httpx2>=2.12.0,<3.0.0` to mirror anthropic's own bound, and extend the comment to state why the anthropic major is capped.
3. **[Axis 3] verdict_tool.py's malformed-response and unmapped-error arms are never exercised** (`src/coder_eval/evaluation/verdict_tool.py:150`) — Routed coverage shows `verdict_tool.py` at 96.49% with lines 150 and 211 missed. Line 150 is the `continue` in `extract_verdict_from_anthropic_response`'s non-dict-block guard:

    149         if not isinstance(block, dict):
    150             continue

and line 211 is the unmapped-validation-error fallback in the message formatter:

    211         messages.append(f"{field} field: {body}" if field else body)

Both are defensive arms on the judge's verdict-parsing path — the first fires on a malformed backend `content` list (a string or `None` element), the second on any `JudgeVerdict` validation error class not explicitly mapped above it. Both are one-line tests: feed `extract_verdict_from_anthropic_response({"content": ["not a dict", <valid tool_use block>]})` and assert the valid block still wins, and feed a `JudgeVerdict` violation outside the mapped set (e.g. an unexpected extra key) and assert the message names the field. Pre-existing (this PR only edited docstrings in this file), and small — filing at Low, not blocking.
4. **[Axis 4] TLS trust anchors for the bearer-token-carrying Bedrock judge POST silently change from the certifi bundle to the OS system trust store** (`src/coder_eval/evaluation/judge_bedrock.py:93`) — `judge_bedrock.py:93` is `async with httpx2.AsyncClient() as client:` — a default client, so it inherits the library's `verify=True` default, and that default's meaning changed under this bump. Verified in the worktree: `httpx 0.28.1`'s `create_ssl_context` does `import certifi` / `ctx = ssl.create_default_context(cafile=certifi.where())`, whereas `httpx2 2.12.0`'s does `import truststore` / `# Default case: rely on the system trust store via truststore.` / `ctx = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)` (`uv run python -c` on `httpx2.create_ssl_context()` returns `<class 'truststore._api.SSLContext'>`, `verify_mode 2`, `check_hostname True`). Verification is still ON — this is not a bypass — but the trusted-root SET for the request at `judge_bedrock.py:98` that carries `judge_bedrock.py:85` `"Authorization": f"Bearer {route.bearer_token}"` is now whatever roots the host/container OS trusts, including any admin- or user-installed enterprise MITM root that certifi would previously have rejected. `trust_env=True` is also the default, so `HTTPS_PROXY` still applies. Two consequences the PR does not state: the `[tool.uv] constraint-dependencies` block (`pyproject.toml:152-163`, `# Fix known CVEs in transitive dependencies`) can no longer pin the judge path's CA bundle via `certifi`, and inside `docker/Dockerfile` (which always bakes `--extra antigravity`, pulling `httpx 0.28.1` alongside `httpx2`) two HTTP stacks now verify against two different anchor sets in one process. Fix: extend the 3-line justification comment at `pyproject.toml:38-40` to state the certifi -> OS-trust-store change, and if the judge path must keep a fixed anchor set, pass an explicit context — `httpx2.AsyncClient(verify=ssl.create_default_context(cafile=certifi.where()))` — rather than relying on the new default. CVSS:3.1/AV:N/AC:H/PR:H/UI:N/S:U/C:L/I:N/A:N

## What's Missing

**Parallel paths:**
- 🟠 The bump changed only the Direct arm's fate: `judge_bedrock.py:78` still puts `"temperature": temperature` in the raw Bedrock JSON body (legal there), while the same value at `judge_anthropic.py:54` can no longer reach the API — yet `criteria/llm_judge.py:210` and `:221` feed both arms the identical `criterion.temperature`. Whichever resolution is chosen (drop the field, route it through `output_config`/`extra_body`, or make it Bedrock-only), both arms must move together or the two judge backends stop being the same measuring instrument. _(trigger: src/coder_eval/evaluation/judge_bedrock.py)_ _(restates: Axis 1: anthropic 1.0.0 dropped `temperature` from `messages.create`)_
- 🔵 The judge is not the only `anthropic` SDK consumer: `errors/categorization.py:60` does `isinstance` dispatch on `AuthenticationError`/`RateLimitError` and `utils.py:487` reads `anthropic.__version__`. I verified both still exist in 1.0.0, so no code change is needed — but the PR records no such check and no test pins that branch, so the next major silently demotes AGENT_RATE_LIMIT / AGENT_AUTH_ERROR to string matching. _(trigger: pyproject.toml)_

**Tests:**
- 🟠 No test binds the real `httpx2` client surface. `_make_async_client` (tests/test_judge_bedrock.py:52-58) hand-rolls a `MagicMock` with `__aenter__`/`post` attached, installed via `monkeypatch.setattr(judge_bedrock.httpx2, "AsyncClient", lambda: ...)` — a zero-arg lambda — so the newly-added direct dependency's `AsyncClient(...)`/`post(url, headers=, json=, timeout=)` surface is never exercised, and the Axis-4 recommended `AsyncClient(verify=...)` fix would break all eight tests on the lambda arity rather than on behaviour. Add one test that binds `inspect.signature(httpx2.AsyncClient.post)` against the kwargs used (or constructs a real client with a mock transport). _(trigger: tests/test_judge_bedrock.py)_
- 🟠 The same blind spot on the Anthropic side: no test in the suite binds `AsyncMessages.create`'s real signature, which is the sole reason a fatal kwarg break shipped with 16 green tests. A `spec=`-bound mock or an `inspect.signature(...).bind_partial(**kwargs)` assertion is the missing test, and it needs no credentials. _(trigger: tests/test_judge_anthropic.py)_ _(restates: Axis 3: Spec-less MagicMock Anthropic client lets the judge tests assert a call shape the real SDK rejects)_
- 🟡 Nothing tests the SDK-dump contract this PR silently re-baselined: `extract_verdict_from_anthropic_response` and `judge_usage.token_usage_from_anthropic_dict` parse `Message.model_dump()`, but every test feeds a hand-written dict. I verified 1.0.0 keeps `content[].{type,name,input}` and `usage.{input_tokens,output_tokens,cache_creation_input_tokens,cache_read_input_tokens}` (new `stop_details`/`output_tokens_details`/`server_tool_use` are additive), so nothing breaks today — a golden fixture generated from a real `Message.model_dump()` is what would make the next bump safe instead of lucky. _(trigger: src/coder_eval/evaluation/verdict_tool.py)_

**Downstream consumers:**
- 🟡 `temperature` is a user-authored, documented field in four places that this PR does not touch: `models/criteria.py:1344`, `docs/TASK_DEFINITION_GUIDE.md:1123` and `:1140`, and `plugins/coder-eval/reference/criteria.md:213` — the last is GENERATED and CE033-guarded, so any resolution also needs `make plugin-reference`. Right now all four promise a knob that the default backend cannot accept. _(trigger: src/coder_eval/evaluation/judge_anthropic.py)_ _(restates: Axis 1: anthropic 1.0.0 dropped `temperature` from `messages.create`)_
- 🔵 `httpx2` is a new DIRECT runtime dependency but CLAUDE.md:322 ("Runtime (always)": …anthropic, claude-agent-sdk, anyio, radon…) was not extended, and no lint rule enforces pyproject↔doc dependency parity (nothing under `tests/lint/` reads the `[project] dependencies` block), so the drift is unguarded in both directions. _(trigger: pyproject.toml)_
- 🟡 The wheel's install constraints changed for every downstream consumer (`coder-eval-uipath`/eval-runner, `action.yml` consumers, both Docker images): a hard `anthropic>=1.0.0,<2.0.0` floor plus a second HTTP stack. Correction to the related Axis-4 note: the dual stack is not extra-specific — `claude-agent-sdk` requires `mcp`, and `mcp` (uv.lock:1137) requires the OLD `httpx` 0.28.1, so the BASE closure now carries httpx (certifi anchors) and httpx2 (OS trust store) in one process. The PR body states none of this. _(trigger: uv.lock)_ _(restates: Axis 4: TLS trust anchors for the bearer-token-carrying Bedrock judge POST change to the OS trust store)_

**Display & mapping dicts:**
- 🔵 `utils.py::get_version_info` stamps `anthropic`, `openai` and `pydantic` versions into every run record's provenance but was not extended with `httpx2`/`httpcore2`. After this PR the HTTP stack is a direct dependency that decides judge retry behaviour and TLS anchors, so a run record can no longer say which stack graded it. _(trigger: pyproject.toml)_

**Daily/nightly:**
- 🟠 Post-merge pipeline impact is unstated and concrete: `docker-publish.yml` runs on every push to main and builds via `uv export --frozen | uv pip install` behind the same safe-chain min-age gate whose exclusion lists (docker/Dockerfile:68, docker/Dockerfile.runtime:56) do not name `anthropic`, so the ghcr `:latest`/`:sha-*` image build is blocked on the org pool until 1.0.0 ages out — as are release.yml and publish-testpypi.yml. The PR should say the merge is timed to the age window rather than letting main go red. _(trigger: uv.lock)_ _(restates: Axis 4: safe-chain minimum-package-age gate blocks anthropic@1.0.0)_
- 🟡 No scheduled job exercises either judge backend: the repo's only cron-driven paid run, `verify-published-action.yml` (`cron: "17 6 * * *"`), drives an inline `file_exists` task with no `llm_judge` criterion, and the burn-in file no job selects is the only real-SDK gate. Consequence worth stating in the PR: the Direct-backend break surfaces first in a user's or the production pipeline's judged run, as a plausible-looking 0.0 score. _(trigger: tests/test_judge_burn_in_live.py)_ _(restates: Axis 3: tests/test_judge_burn_in_live.py is selected by no CI step)_
- 🔵 The production `--backend bedrock` judge path changes behaviour inside the container the daily pipeline runs: the bearer-token POST at judge_bedrock.py:93-98 now verifies against the image's OS trust store instead of the certifi bundle, and the `[tool.uv] constraint-dependencies` CVE-pin block (pyproject.toml:152-163) can no longer pin that anchor set. The PR should state the expected effect on the nightly/production judge, or pin an explicit SSL context. _(trigger: src/coder_eval/evaluation/judge_bedrock.py)_ _(restates: Axis 4: TLS trust anchors for the bearer-token-carrying Bedrock judge POST change to the OS trust store)_

- [pyright] Set `enableTypeIgnoreComments = false` in `[tool.pyright]` (pyproject.toml, beside `typeCheckingMode = "standard"`). The repo type-checks with pyright ONLY, but carries 30 mypy-style `# type: ignore[code]` comments in `src/` (vs 5 `# pyright: ignore[...]`); pyright ignores the bracketed code and treats each as a BLANKET line suppression that kills any diagnostic whose RANGE covers that line. MEASURED in the PR worktree (generated a config from `[tool.pyright]` with only this key flipped, ran `.venv/bin/pyright --outputjson`): 22 errors / 3 warnings tree-wide, one of which is `src/coder_eval/evaluation/judge_anthropic.py:49 reportCallIssue — No overloads for "create" match the provided arguments`, i.e. exactly the shipped critical bug. The remaining 21 are pre-existing and bounded (6 `reportIncompatibleVariableOverride` on `agent_config.py` discriminator re-declarations, 3 optional-`openai_codex` `reportMissingImports`, 2 `LoggerAdapter` type args, plus 10 singletons) — a one-pass migration to rule-scoped `# pyright: ignore[<rule>]`, not a refactor. Also apply it to the second pass automatically: `tests/lint/pyright_config.py::build_config` copies every rule setting verbatim from `[tool.pyright]`, so the flip covers both passes with no twin to drift. _Prevents:_ Critical: `temperature=` passed to `AsyncMessages.create` after anthropic 1.0.0 removed it (judge_anthropic.py:54) — silently scoring every Direct-backend `llm_judge` row 0.0. High: the blanket `# type: ignore[arg-type]` at judge_anthropic.py:55 that swallowed the whole-call `reportCallIssue` and kept `make typecheck` green.
- [ce-lint] CE068 — `# type: ignore` is banned in `src/`; use `# pyright: ignore[<rule>]`. A `BaseRule` in `tests/lint/rules/ce068_no_mypy_type_ignore.py`, wired into `ALL_RULES` in `tests/lint/runner.py`, plus a `"CE068"` entry in `[tool.ruff.lint] external` (that list already carries CE037/CE038/CE039). Detection is a comment scan, not AST: flag any `# type: ignore` token in a `src/coder_eval/**.py` line. It is the readability half of the pyright flip above — with `enableTypeIgnoreComments = false` a new `# type: ignore` suppresses nothing and the error simply reappears, which is confusing rather than instructive; CE068 names the correct spelling at the point of the mistake and keeps the rule-scoped form (which pyright honours per-rule, verified: with `# pyright: ignore[reportArgumentType]` on the `tools=` line the `temperature` `reportCallIssue` STILL fires) the only way to suppress. _Prevents:_ Recurrence of the judge_anthropic.py:55 over-broad suppression, and the other 29 `# type: ignore` sites in `src/` that are each silently a whole-line blanket today.
- [ruff] Enable `ICN` and add a banned-alias declaration:

```toml
[tool.ruff.lint]
select = ["E","F","I","N","W","UP","B","SIM","RUF","PLR0915","PLR0912","ICN"]

[tool.ruff.lint.flake8-import-conventions.banned-aliases]
httpx2 = ["_httpx", "httpx"]
```

MEASURED with an isolated config over `src` + `tests`: exactly 2 hits, `tests/test_judge_bedrock.py:152` and `:218` (`ICN002 \`httpx2\` should not be imported as \`_httpx\``) — zero collateral. Rejected the adjacent candidate: `PLC0415` (import-outside-top-level) would also flag those two function-local imports, but it fires 1191 times across `src` + `tests`, so it is not a viable gate here. _Prevents:_ Low: `import httpx2 as _httpx` in two duplicated function-local imports, which makes the new library read as the old one in the very PR that migrates off `httpx`.
- [ce-lint] CE069 — every `tests/**/*_live.py` must be selected by at least one CI pytest invocation. A `@pytest.mark.lint` class (it reasons over YAML + a file glob, not one `.py` AST — same filing as `tests/lint/workflow_outputs.py`, the CE035 workflow reader), living in `tests/lint_tests/` with its detection body beside `tests/lint/workflow_outputs.py`. Rule: for each file matching `tests/*_live.py` (or carrying `pytest.mark.live`), require its path to appear in a `run:` line of some `.github/workflows/**.yml` step that invokes pytest, OR an `EXEMPT` entry naming the reason (the CE030 doc-parity precedent). CURRENT STATE in the PR base: 5 live files, 3 selected (`test_claude_settings_enforcement_live.py` at pr-checks.yml:723/735, `test_codex_agent_live.py`:864, `test_byoa_plugin_live.py`:941) and 2 selected by nothing — `test_judge_burn_in_live.py` and `test_docker_workdir_live.py`. `make test`/`make verify` run `-m "not live"`, so a live file wired to nothing is dead weight that still reads as coverage; `test_judge_burn_in_live.py`'s own docstring even claims it is "safe to run in CI". _Prevents:_ High: `tests/test_judge_burn_in_live.py` selected by no CI step, which is the direct reason the anthropic 1.0.0 bump had no real-SDK gate — its `test_llm_judge_anthropic_direct_tool_channel` is the one test that would have caught the removed kwarg.
- [ce-lint] CE070 — every direct runtime dependency states its upper-bound decision. A `@pytest.mark.lint` class reading `[project].dependencies` from `pyproject.toml` with `tomllib`, plus a line scan for the adjacent comment block: (a) an entry with an upper bound (`<N.0.0`) must have a comment above it that mentions the capped distribution name; (b) an entry with NO upper bound must carry a `# uncapped:` rationale comment. The block already follows this convention voluntarily — pyproject.toml:47-51 explains both `opentelemetry-sdk<2.0.0` and `azure-monitor-opentelemetry-exporter<1.1.0` — and this PR is where it broke in both directions: line 37 `anthropic>=1.0.0,<2.0.0` gained a cap with no rationale (while this very PR is the proof that an anthropic major removes call kwargs), and line 41 `httpx2>=2.12.0` shipped uncapped even though the comment above it argues httpx2 is coupled to anthropic's major (`importlib.metadata.requires('anthropic')` pins `httpx2<3,>=2.0.0`). _Prevents:_ Low: the internally inconsistent dependency hunk at pyproject.toml:37-41 (uncapped `httpx2`, unexplained `anthropic<2.0.0`).
- [ce-lint] CE071 — a mock standing in for a third-party SDK symbol must be `spec=`/`autospec=`-bound. A `BaseRule` in `tests/lint/rules/ce071_specced_sdk_mock.py` over `tests/**.py`: flag `patch("<dotted.path>.<Name>", ...)` / `monkeypatch.setattr(...)` whose replacement is a bare `MagicMock()` / `AsyncMock()` when `<Name>` is imported into the target module from a distribution outside `coder_eval` (resolve via the module's own import list — `AsyncAnthropic`, `AsyncClient`, `ClaudeSDKClient`, `boto3`, …). Blast radius measured: 4 call sites, all in `tests/test_judge_anthropic.py` (lines 53, 71, 83, 102), against `MagicMock(spec=…)`/`autospec=` already used in 7 test files. This is the defence-in-depth layer behind the pyright flip: pyright covers a typed SDK like `anthropic`, but `claude-agent-sdk` / `openai_codex` are optional or untyped, and there pyright sees nothing while a `spec=`-bound double still fails on a renamed kwarg. _Prevents:_ Critical (test-health): `_make_client()`'s bare `MagicMock()` at tests/test_judge_anthropic.py:31 let line 86 positively assert `kwargs["temperature"] == 0.7` for a call the installed SDK rejects — 16 tests green over a fatal break.
- [ce-lint] CE072 — a judge transport must not let a non-API exception reach the criterion decorator. A `BaseRule` over `src/coder_eval/evaluation/judge_*.py`: the `try` that wraps the outbound SDK/HTTP call must have, besides its typed arms (`except APIError`, `except httpx2.HTTPError`), a catch-all `except Exception` that raises `JudgeInfrastructureError`. Today `judge_anthropic.py:48-62` catches `APIError` only, so a `TypeError` from a changed SDK signature falls through to `handle_criterion_errors_async`, whose `_ESCALATING_EXCEPTIONS` is `(JudgeInfrastructureError, CheckerMisuseError)` (criteria/base.py:81) — and is booked as `CriterionResult(score=0.0)`. This is CE039's mirror image one seam upstream: CE039 already forbids a criterion checker from booking an IO/config error as a gating 0.0 for exactly this reason ("a judge-infra outage or a checker-contract misuse is not an agent failure"), and the transport seam is where the same mistake is currently legal. Note the interaction with CE005 (`no_silent_except`): a handler that raises already satisfies CE005, so the two rules agree. _Prevents:_ The CONSEQUENCE half of the critical finding — an SDK break reported as the agent scoring 0.0 across every judged row instead of escalating to `FinalStatus.ERROR`. Also caps the blast radius of the next SDK bump, whatever its shape.
- [ce-lint] CE073 — a credential-carrying HTTP client must pass `verify=` explicitly. A `BaseRule` over `src/coder_eval/**`: flag `httpx2.AsyncClient(...)` / `httpx2.Client(...)` constructed without a `verify=` keyword. One site today, `src/coder_eval/evaluation/judge_bedrock.py:93` (`async with httpx2.AsyncClient() as client:`), which POSTs the `Authorization: Bearer …` header built at :85. The library default is still `verify=True`, but its MEANING changed silently under this bump — verified in the worktree, `httpx 0.28.1`'s `create_ssl_context` builds `ssl.create_default_context(cafile=certifi.where())` while `httpx2 2.12.0` returns a `truststore._api.SSLContext` over the OS trust store. An explicit `verify=` makes the anchor set a source-visible decision that a dependency bump cannot move underneath the judge path. _Prevents:_ Low (security): the undeclared certifi -> OS-trust-store change on the bearer-token-carrying Bedrock judge POST, which also silently removes that path from the `[tool.uv] constraint-dependencies` certifi pin's reach.

- Wire `tests/test_judge_burn_in_live.py` into the existing `live-tests` job in `.github/workflows/pr-checks.yml` (job at :645). That job already holds every credential the file needs at job scope — `ANTHROPIC_API_KEY` (:656), `AWS_BEARER_TOKEN_BEDROCK` (:660), `AWS_REGION` (:661) — and installs the Claude CLI at :678 for the `agent_judge` test. Add a step mirroring :723: `.venv/bin/pytest tests/test_judge_burn_in_live.py -m live -v --tb=short --strict-markers -ra --junit-xml=tmp/junit-burn-in.xml`, and extend the existing "tests actually ran (not silently skipped)" JUnit passed-count parser (:739-770) to that XML so a credential-skip is red rather than green. Cost: 2 real `llm_judge` calls (Sonnet 4.6 / Haiku 4.5 per the file's docstring) plus one `agent_judge` SDK agent, so treat the docstring's "3 real LLM calls" as a floor. _Why not static:_ CE069 can only prove the file is SELECTED by some CI step; it cannot prove the call succeeds against the real SDK and a live endpoint. Confirming that `messages.create` accepts the kwargs we send, that the tool-channel verdict parses, and that Bedrock and Direct still agree needs credentials and network. _Prevents:_ High: the burn-in file being reachable only via `make test-live` (Makefile:55), which `make verify` excludes — the missing gate behind the critical anthropic 1.0.0 break.
- Add an OFFLINE SDK call-contract test (e.g. `tests/test_sdk_call_contracts.py`, no `live` marker) that binds each transport's production kwarg set against the installed SDK's real signature: `inspect.signature(anthropic.resources.messages.AsyncMessages.create).bind(client_self_placeholder, **PRODUCTION_KWARGS)`, and the same for the `claude-agent-sdk` entry points the agents call. Derive `PRODUCTION_KWARGS` from one shared dict that the transport itself splats, so the test and the call site cannot describe different calls. This is the check that fails in 0.5s on `make test` — including in the `no-uipath-extra` (:280) and `windows-smoke` (:405) jobs — the moment a dependency bump renames or drops a parameter. _Why not static:_ The accepted parameter set is a property of the INSTALLED distribution, not of our source tree. pyright reaches it for `anthropic` (which ships `py.typed`) once the blanket ignore is gone, but not for the untyped/optional SDKs (`openai_codex` already resolves to `reportMissingImports` in the type pass), and no static rule can see a signature that changes in `uv.lock` rather than in a diff of `src/`. _Prevents:_ Critical: `temperature=` on `AsyncMessages.create`. Critical (test-health): the same break being asserted GREEN by an unspecced double — a bind-based test cannot be fooled by a mock.
- Add a dependency-major-bump protocol job: a `pull_request`-only step that diffs `pyproject.toml` + `uv.lock` against the base ref and, when a DIRECT dependency's major version moves, requires (a) the `live-tests` job to have actually run on that PR, and (b) a row in a `## Dependency majors` table (CONTRIBUTING.md or docs/) naming the upstream API surfaces re-verified and the call sites checked. Same shape as the repo's other protocol gates — diff-based, `pull_request`-only, demanding a written row rather than guessing. _Why not static:_ A working tree has no base ref, so the trigger ("this PR moves a major") is not computable in `make verify` — it exists only relative to a merge base, which is why this class of gate has to be a CI job rather than a lint rule. The second half is also judgment: which surfaces a major touched cannot be derived from the version string. _Prevents:_ Critical: an `anthropic` 0.102 -> 1.0.0 bump reviewed as a pure `httpx` -> `httpx2` rename, with the removal of `temperature`/`top_p`/`top_k` from the Messages surface unnoticed and unstated.
- Add a minimum-package-age preflight (`make deps-age`, plus one CI step before the install) that reads the `upload-time` fields uv.lock already records for every distribution and reports each newly-locked or newly-bumped entry younger than the org window, naming it. This PR's block was `anthropic@1.0.0` (uv.lock:167, uploaded 1 day before the PR); the httpx2 family (uv.lock:906/:946, 3 days) cleared it. Two `PR Quality Checks` jobs went red with a raw `403 Forbidden - blocked by safe-chain direct download minimum package age (anthropic@1.0.0)` from inside `uv sync`, and the same gate sits on `release.yml:78`, `publish-testpypi.yml:41`, `docker/Dockerfile:68` and `docker/Dockerfile.runtime:56`, so the wheel cannot be built either. A named preflight also makes the right remedy obvious — wait out a window that expires in days, rather than adding `anthropic` to `SAFE_CHAIN_MINIMUM_PACKAGE_AGE_EXCLUSIONS` in five places and permanently waiving the control for the most security-sensitive dependency in the tree. _Why not static:_ The verdict depends on wall-clock time: the identical lockfile is a violation today and clean next week, so it cannot be a stable `make lint` assertion (a lint rule that turns green by itself is worse than none). It has to be evaluated at CI/install time, and it has to run BEFORE the install step so the failure names the distribution instead of surfacing as an opaque 403 inside `uv sync`. _Prevents:_ Medium (security): the safe-chain min-age block on `anthropic@1.0.0`, currently visible only as a 403 inside a failing install, with the blast radius (every packaging path) unstated in the PR.
- Pin the judge path's TLS anchor set with a runtime assertion rather than a comment: a unit test that constructs the judge client exactly as `judge_bedrock.py` does and asserts the SSL context's provenance is the intended one (`ssl.SSLContext` from `certifi` vs `truststore._api.SSLContext`), so a future library default flip is a red test instead of a silent change of who can MITM a bearer token. Pairs with CE073, which forces `verify=` to be written down. _Why not static:_ The defect is a change in the DEFAULT VALUE's meaning inside an installed library — `verify=True` reads identically before and after the bump, and no rule over our source can tell that `create_ssl_context` switched from `certifi.where()` to `truststore`. It needs the library imported and the context object inspected. _Prevents:_ Low (security): the certifi -> OS-trust-store change on the bearer-token-carrying Bedrock judge POST, and the two-anchor-set situation inside `docker/Dockerfile` where `httpx 0.28.1` (via `--extra antigravity`) and `httpx2` coexist in one process.
- Cover the judge verdict parser's defensive arms and keep them covered: add the two one-line tests for `src/coder_eval/evaluation/verdict_tool.py:150` (a non-dict element in a backend `content` list — assert a valid `tool_use` block still wins) and `:211` (a `JudgeVerdict` validation error class outside the mapped set — assert the message names the field), and consider a per-package coverage floor for `evaluation/` so the parsing seam cannot drift below it while the global 80% gate stays green. _Why not static:_ The gap is unexecuted BEHAVIOUR on malformed input, which is a runtime measurement; a static rule can see that a branch exists but not that nothing ever entered it. _Prevents:_ Low: the two uncovered defensive arms in `verdict_tool.py` — the malformed-response guard and the unmapped-validation-error fallback, both on the judge path this PR touches.

## Top 5 Priority Actions

1. Drop `temperature=temperature` from the `client.messages.create(...)` call at src/coder_eval/evaluation/judge_anthropic.py:54 (anthropic 1.0.0 removed sampling params from the Messages surface) and decide explicitly whether `LLMJudgeCriterion.temperature` becomes Bedrock-only or is forwarded via `extra_body` — today every `llm_judge` criterion on the default Direct backend scores the agent 0.0 for a harness bug.
2. Stop scoring harness faults as agent failures: extend `_ESCALATING_EXCEPTIONS` at src/coder_eval/criteria/base.py:81 (or wrap the SDK call so a signature/contract break raises `JudgeInfrastructureError`) so a `TypeError` inside a judge escalates to `FinalStatus.ERROR` instead of being captured into a `score=0.0` `CriterionResult` while the run finishes green.
3. Rebuild the judge test double as a signature-honest mock — `MagicMock(spec=AsyncMessages)` or an `inspect.signature(...).bind(...)` assertion at tests/test_judge_anthropic.py:31 — and delete the now-false `assert kwargs["temperature"] == 0.7` at tests/test_judge_anthropic.py:86, so a removed or renamed SDK kwarg fails the unit suite instead of passing it.
4. Replace the blanket `# type: ignore[arg-type]` at src/coder_eval/evaluation/judge_anthropic.py:55 with the rule-scoped `# pyright: ignore[reportArgumentType]` (verified to re-expose the `reportCallIssue` for the bad kwarg), and set `enableTypeIgnoreComments = false` or add a CE rule, since all ~30 mypy-style ignores in `src/` are silently whole-diagnostic suppressions under this pyright-only setup.
5. Wire tests/test_judge_burn_in_live.py:50 into the existing `live-tests` job in .github/workflows/pr-checks.yml:645 (it already holds both credential sets and the JUnit passed-count parser at :739) so the only real-SDK judge gate actually runs and a silent credential-skip fails CI.

---

**Stats:** 2 🔴 · 2 🟠 · 1 🟡 · 4 🔵 across 8 axes reviewed.

uipreliga (approved):
Fix what you agree with and 🚢

PR #142, opened 2026-08-27
feat(docker): bake litellm extra into the docker image

akshaylive:
## Summary
- Bakes the `litellm` extra into the docker image alongside `codex` and `antigravity`
- `litellm` backs the `checker_context.api_route.route: litellm` judge dispatch — needed in-container since the checker runs there too under `DockerRunner` isolation, not just on the host
- All three extras resolve from public PyPI per `uv.lock`, so no private-index credentials are needed

## Test plan
- [x] `make test` passes (4367 passed, 2 skipped) before and after rebase onto latest `main`


uipreliga (approved):
🚢

PR #152, opened 2026-09-02
perf(evalboard): cut page load time and show loading state while pages load

bai-uipath:
Loading the front page, filtering to a tag inside a run, and the download button were all slow. Reviewing the result turned up more on the same pages, so this also cuts the HTML those pages ship, gives every slow route a loading state, restores a filter segment that had gone missing, and corrects a model count in the run header.

## Rough before and after

| What | Before | After |
|---|---|---|
| Front page, cache warm | 1.25s | under 10ms, and warm now lasts a day on settled runs instead of 5 minutes |
| Front page, cold render | ~205 run.json files, ~420 MB read | ~294 MB of that no longer read at all |
| Ad-hoc section | every candidate loaded, 165 runs | newest page only, ~20 runs |
| Run page, per request | run.json parsed twice, plus a serial 20-run scan | parsed once, scan batched; under 10ms warm |
| Tag click inside a run | full server re-render of a force-dynamic route | no request at all |
| Mature-source scan | up to 20 sequential reads | 4 concurrent batches |
| Whole-run download | ~10k blobs, ~400 MB, buffered in memory before the first byte | removed |
| Per-task zip checksum | 94ms per 50 MB, on the event loop | 2ms |
| Run page HTML | 7.6 MB raw, 222 KB gzipped | 6.0 MB raw, 205 KB gzipped |
| /trends HTML | 5.6 MB raw | 4.5 MB raw |
| Feedback while a page loads | none, the previous page just sits there | skeleton immediately |

Two honesty notes on that table. The warm figures are local; the "before" warm and cold figures are production App Insights, and the production cold path stays unverified until this deploys. And the HTML reduction is mostly a parse win rather than a transfer win, since gzip already collapses repeated class text.

## What changed

**Front page.** A cold render read about 205 run.json files off the `/home` Azure Files mount. Four independent causes: the ad-hoc section loaded every candidate ever uploaded in order to show ten rows; per-run projections expired every 5 minutes however old the run was; the same run.json was read and parsed twice per request; and the mature-source scan walked earlier runs one at a time. All four are now bounded or shared.

**Filtering inside a run.** Every tag click re-ran the whole server component to return markup the browser then recomputed anyway, since nothing on that page reads the filter params server-side. Filters now commit through the native History API: the URL stays shareable, the back button still works, and there is no round-trip.

**Download.** Whole-run download is removed. A nightly run is about 10k blobs and 400 MB, and zipping it fetched every blob with no concurrency cap and buffered the whole archive in memory before sending a byte. Per-task download stays, and its checksum now comes from `zlib` rather than a per-byte loop.

**Loading states.** Every heavy route is force-dynamic, so a navigation has to run the server component before it can render anything, and there was no Suspense boundary anywhere in the app. The browser held the previous page for the whole wait with no sign the click had registered. There is now a route-level skeleton for the front page (covering trends, watchlist, path-to-ga, scribe and runs/latest by nesting), the run page, and the task page. The run page's own `Suspense fallback={null}` is replaced with the same body skeleton, since it had been leaving a blank hole under an already-rendered header. Blocks mirror the real breakpoints, so the layout does not move when content lands, and tests assert no fixed width can overflow a 320px viewport.

**Payload on the run page and /trends.** 55% of the markup on both pages was `class` attribute text, and a single 130-character chip string accounted for 1.4 MB on its own. Those combinations now live in the stylesheet, which the browser parses once and reuses across navigations. No visual change: every generated rule was diffed against the utilities it replaced in the built CSS. The same audit found every other page already under 610 KB, so those two were the only ones worth touching.

**Harness filter.** Segments were discovered from the last 12 runs, so delegate vanished between its weekly firings and read as a harness that had been removed. The known set now always gets a segment, and discovery still drives the chart series.

**Model count in the run header.** The header claimed "+1 more" on runs that used one model end to end, because a task that errors before the model resolves records the qualified id while completed tasks record the bare one. The count now groups on the normalized id, the same key pricing already resolves on. Display still shows what the run actually recorded.

## On caching

A run older than a day now holds its projection for a day instead of 5 minutes. That reads like a staleness risk and is not, because the on-disk blob cache underneath never re-fetches a file once it exists: a sidecar edited in storage was already stale until someone pressed Refresh, whatever TTL sat above it. Refresh is still the only thing that clears it, and it now evicts the projection alongside the disk copy, which is the part that actually needed fixing. Deleting or re-uploading a run behaves exactly as before. The one assumption worth recording is single-instance, since tag eviction is per process; the site runs one worker and has no autoscale rule.

## Also done outside the repo

`alwaysOn` was `false` on the App Service and is now `true` (`flow-evalboard` / `rg-coder-eval-tests`). It is free on the current P2v3 plan and was the last cause of a cold front page that no code change could reach, since the worker idled out and the first visitor after a quiet stretch paid the whole boot. No repo change was needed and none is needed to keep it: nothing in the deploy path writes `siteConfig` and neither repo carries IaC. Verified healthy afterwards, 30 requests in the next five-minute bin, all 200.

Two notes for anyone checking by hand. An anonymous `curl` against the public hostname answers 401 from Easy Auth, which is expected and not a failed boot. `az webapp auth show` reports nulls because the site is on classic V1 while the CLI extension reads V2.

## Verification

`tsc --noEmit`, 637 vitest tests, and `next build` all pass. New coverage for the id-date extraction, the batched mature-source walk, the refresh tag eviction, the zip checksum, the harness segments, the model tally, and the skeletons (accessible name, reduced motion, no overflowing width, both grid breakpoints present).

### DIFF
--- a/evalboard/app/_components/__tests__/harness-selector.test.tsx
+++ b/evalboard/app/_components/__tests__/harness-selector.test.tsx
@@ -62,6 +62,57 @@ describe("HarnessSelector", () => {
         ).toHaveAttribute("aria-pressed", "true");
     });
 
+    test("a known harness absent from the discovery window still shows", () => {
+        // Delegate runs weekly, so it falls out of the 12-run discovery window
+        // between firings. The segment must survive that, or the filter reads as
+        // "delegate was removed" rather than "delegate hasn't run lately".
+        render(
+            <HarnessSelector
+                current={null}
+                harnesses={["claude-code"]}
+                includeAll
+            />,
+        );
+        for (const h of KNOWN_HARNESSES) {
+            expect(
+                screen.getByRole("button", { name: harnessShortLabel(h) }),
+            ).toBeInTheDocument();
+        }
+    });
+
+    test("a discovered newcomer is kept, and sorts after the known set", () => {
+        render(
+            <HarnessSelector
+                current={null}
+                harnesses={["zzz-new-harness"]}
+                includeAll
+            />,
+        );
+        const names = screen
+            .getAllByRole("button")
+            .map((b) => b.getAttribute("aria-label"));
+        expect(names).toContain(harnessShortLabel("zzz-new-harness"));
+        // Known harnesses hold display order; the newcomer lands last.
+        expect(names[names.length - 1]).toBe(
+            harnessShortLabel("zzz-new-harness"),
+        );
+    });
+
+    test("no segment is duplicated when discovery and the known set overlap", () => {
+        render(
+            <HarnessSelector
+                current="codex"
+                harnesses={[...KNOWN_HARNESSES]}
+                includeAll
+            />,
+        );
+        const labels = screen
+            .getAllByRole("button")
+            .map((b) => b.getAttribute("aria-label"))
+            .filter((n): n is string => n != null);
+        expect(new Set(labels).size).toBe(labels.length);
+    });
+
     test("every segment is named for screen readers, not color-only", () => {
         render(
             <HarnessSelector
--- a/evalboard/app/_components/__tests__/skeleton.test.tsx
+++ b/evalboard/app/_components/__tests__/skeleton.test.tsx
@@ -1,0 +1,97 @@
+import { describe, expect, test } from "vitest";
+import { render } from "@testing-library/react";
+import RootLoading from "@/app/loading";
+import RunLoading from "@/app/runs/[id]/loading";
+import TaskLoading from "@/app/runs/[id]/[...task]/loading";
+
+const CASES = [
+    ["root", RootLoading, "Loading page"],
+    ["run", RunLoading, "Loading run"],
+    ["task", TaskLoading, "Loading task"],
+] as const;
+
+// Every class on every rendered node, flattened into tokens.
+function classTokens(root: HTMLElement): string[] {
+    const out: string[] = [];
+    for (const el of root.querySelectorAll<HTMLElement>("*")) {
+        out.push(...el.className.split(/\s+/).filter(Boolean));
+    }
+    return out;
+}
+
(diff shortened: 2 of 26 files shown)


bai-uipath:
Validated the ad-hoc change against the real `runs` container rather than reasoning about it. Pulled `run.json` last-modified for all 164 ad-hoc candidates and compared true recency against what the new code would actually load:

```
candidates=164   undated=2   loaded_by_new_code=22   (was 164)
read reduction: 164 -> 22   (87% fewer run.json reads)

TRUE newest 10 (by run.json lastModified) — all inside the loaded set? YES
   1. ok  2026-09-02 22:38  adhoc-2026-09-02_21-59-13
   2. ok  2026-09-02 21:45  adhoc-2026-09-02_15-08-28
   3. ok  2026-09-02 19:26  adhoc-2026-09-02_18-30-39
   ...
  10. ok  2026-09-01 18:50  adhoc-2026-09-01_14-03-30

First true-rank NOT loaded: #21 (so the section is exact through rank 20)
```

The section renders 10 rows and is exact through 20, so there is a full page of headroom. The guarantee scales with the limit too — expanding via `alimit` raises the budget by the same amount, so "Show more" stays exact rather than degrading as you page back.

Two notes for review:

- Only **2** ids in the whole container carry no date (`sdk-live-r2-final`, `adhoc-flow-v2-preview-20260820`). Those are always loaded regardless of budget, so they can't be ordered out by a key they don't have. `deploys/` is the third non-run prefix; it has no `run.json` at all and drops out as an unreadable overview, which is what `ADHOC_LOAD_SLACK` absorbs.
- Largest id-date vs upload-time skew is ~142h, on runs re-uploaded well after they started. That doesn't affect ordering: the rows sort on `start_time` read from run.json, which travels with the content, so a late re-upload keeps its original position either way. The id date only decides what gets read.

uipreliga (approved):
🚢

PR #155, opened 2026-09-03
fix(pricing): refresh the rate card and add gemini 3.7/3.8 Flash

bai-uipath:
Re-verified every row of the built-in rate card against the vendor cards, and added the models the antigravity bump now needs. Four rows were stale; three models were missing entirely.

## Why now

`ANTIGRAVITY_MODEL` was just moved to `gemini-3.8-flash` in both variable stores (the `skills` repo variable and the ADO `coder-eval-athena-config` group). Antigravity computes cost locally from tokens against this table, so an unpriced id books no cost at all — the run still passes, but reports `cost_complete: false` and a null bill. That prompted a full audit rather than just the one row.

## Repriced

| model | was | now |
|---|---|---|
| `claude-sonnet-5` | $3 / $15 | **$2 / $10** |
| `gpt-5.6-sol` | $5 / $30 | **$4 / $20** |
| `z-ai/glm-5.2` | 0.7168 / 2.2528 | **0.966 / 3.036** |
| `deepseek/deepseek-v4-pro` | 0.435 / 0.87 | **1.0308 / 2.0616** |

`claude-sonnet-5` is the significant one. The table deliberately held $3/$15 against the $2/$10 introductory rate lapsing on 2026-08-31. Anthropic has since made $2/$10 the standard price and cancelled the scheduled increase, so that hedge was overstating every run of the nightly's own model by 50%.

The two OpenRouter rows come from the live `/api/v1/models` feed. Both had drifted up, and `deepseek-v4-pro`'s cache-read rate was low by roughly 24x.

## Added

- `gemini-3.8-flash` and `gemini-3.7-flash` — same list rate as 3.6.
- `claude-fable-5-1` — it prices cache hits at 0.025x input, not the 0.1x every other Claude model uses. Worth a row of its own so nobody copies Fable 5's figure onto it.

## Left alone, deliberately

The three Bedrock open-weight rows (`deepseek.v3.2`, `zai.glm-5`, `moonshotai.kimi-k2.5`) are unchanged. AWS does not publish eu-north-1 figures for them, and the existing comment warns against substituting the US column. Carried forward and documented rather than guessed.

## Verification

Full Python suite, evalboard `tsc --noEmit`, and all 608 evalboard tests pass, including the `pricing.py` ↔ `pricing.ts` parity guard. The 13 failures in `test_judge_litellm.py` / `test_reports_stats_nonfinite.py` on my machine reproduce identically with these changes stashed — a local venv artifact, not this branch.

The gemini rows were confirmed end to end: an antigravity run on `gemini-3.8-flash` went from `cost_complete: false` / null cost to `cost_complete: true` / \$0.0685 once the row was present.

### DIFF
--- a/evalboard/lib/pricing.ts
+++ b/evalboard/lib/pricing.ts
@@ -24,8 +24,8 @@ export const PRICING: Record<string, Pricing> = {
     // alias must never inherit the older tier. Undated aliases each need their
     // own key: resolvePricing's fallback only strips a trailing date (dated →
     // undated), it cannot invent one.
-    // Fable 5.1 prices cache HITS at 0.025x input ($0.25), not the 0.1x every
-    // other Claude model uses — Fable 5 pays $1 on the identical $10 base.
+    // Fable 5.1 prices cache hits at 0.025x input, not the 0.1x every other
+    // Claude model uses. Fable 5 pays $1 on the identical $10 base.
     "claude-fable-5-1": p(10, 50, 12.5, 0.25),
     "claude-fable-5": p(10, 50, 12.5, 1),
     "claude-opus-5": p(5, 25, 6.25, 0.5),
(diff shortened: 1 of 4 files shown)


uipreliga (approved):
🚢

PR #156, opened 2026-09-04
feat(orchestrator): tell headless runs that no user is present

rockymadden:
## Problem

Task YAMLs carry hand-copied lines like *"Do NOT ask for approval, confirmation, or feedback"* to get autonomous behavior. That phrasing forbids **asking**. It does not say **nobody is there to ask**.

Those are different instructions, and an agent can honor the first while still stopping at a consent gate to wait for a reply that never arrives. That is not hypothetical: in the 2026-09-04 `skill-flow-*` nightly, 5 of 8 tasks built and validated a flow, reported success, and never executed it. Every one of those prompts contained the "Do NOT ask" line.

Measured across the `uipath-maestro-flow` suite (128 tasks):

| | count |
|---|---|
| Tasks stating the run is headless | **0** |
| Non-simulated tasks | 119 |
| ...saying nothing about autonomy | **51** |
| ...saying something, across **8** wording variants | 68 |
| Silent **and** graded on a real run of the artifact | **9** |

The instruction is hand-maintained, inconsistent, and missing from nearly half the suite.

## Fix

The orchestrator already knows which case it is in. Simulated tasks are routed to `_simulation_dialog_loop` at `orchestrator.py:1925` and never reach the prompt built in `_evaluation_loop`. The two paths are separate functions, so stating it once at that construction site covers every headless task, cannot reach a task that has a live user, and needs no conditional.

```python
HEADLESS_PREAMBLE = (
    "This run is headless. No user is present, and nobody will answer a question "
    "or grant an approval. Do not wait for input: choose the best available option "
    "and record every decision and assumption in your final response. Stop only "
    "when proceeding would be unsafe or irreversible, and say what you stopped on. "
    "Instructions in the task below take precedence over this paragraph."
)
```

Ordering is cwd → preamble → task prompt.

## Why no opt-out flag

The last sentence defers to the task. A task that deliberately asserts a refuse-and-surface still overrides it. `skill-flow-eval-no-auto-upload` is the live example: it instructs the agent to refuse an upload and surface the choice, and it keeps working without anyone remembering to set a flag.

I would rather find out that is too soft from a real run than add a config field on speculation.

## Tests

Two, both in `tests/test_orchestrator.py`:

- `test_evaluation_loop_prepends_headless_preamble` — the preamble reaches the agent, and the ordering holds.
- `test_simulated_task_bypasses_the_headless_prompt_path` — a simulated task delegates to the dialog loop and never builds the headless prompt.

The second is the one that matters. The design rests entirely on the function boundary, so a refactor that merges the loops should fail loudly.

## Verification

- `ruff format --check` / `ruff check` on `src` and `tests`: clean
- `pyright` on both changed files: 0 errors
- `make lint` (custom CE rules): 388 passed
- 450 orchestrator, run-limit, simulation, early-stop and error-handling tests pass

## Downstream

`UiPath/skills` pins this via `tests/.coder-eval-version` (currently 0.11.5). Once this releases, that pin bumps and all 119 headless flow tasks get the declaration. The 68 hand-copied autonomy lines become redundant and can be swept separately.

### DIFF
--- a/src/coder_eval/orchestrator.py
+++ b/src/coder_eval/orchestrator.py
@@ -83,6 +83,19 @@ logger = logging.getLogger(__name__)
 # wins the race against the asyncio cancel path (which doesn't).
 _WAIT_FOR_GRACE_SECONDS = 2.0
 
+# Prepended to the prompt on the non-simulated path only. Task YAMLs have been
+# carrying hand-copied "Do NOT ask for approval" lines to get the same effect,
+# which forbids asking without saying nobody is there to ask — an agent can honor
+# it and still stop at a consent gate waiting for a reply that never comes. The
+# simulated path (_simulation_dialog_loop) has a live user, so it never sees this.
+HEADLESS_PREAMBLE = (
+    "This run is headless. No user is present, and nobody will answer a question "
+    "or grant an approval. Do not wait for input: choose the best available option "
+    "and record every decision and assumption in your final response. Stop only "
+    "when proceeding would be unsafe or irreversible, and say what you stopped on. "
+    "Instructions in the task below take precedence over this paragraph."
+)
+
 
 def _close_subprocess_transport(proc: asyncio.subprocess.Process | None) -> None:
     """Release a finished subprocess's pipe transport deterministically.
@@ -1936,7 +1949,9 @@ class Orchestrator:
         self.result.iteration_count = iteration
 
         # Communicate with agent (with retry logic)
-        prompt_with_cwd = f"Your working directory is: {sandbox_dir.resolve()}\n\n{current_prompt}"
+        prompt_with_cwd = (
+            f"Your working directory is: {sandbox_dir.resolve()}\n\n{HEADLESS_PREAMBLE}\n\n{current_prompt}"
+        )
         logger.debug(f"Sending prompt: {current_prompt[:100]}...")
 
         # The agent owns its lifecycle events now (AgentStartEvent fires inside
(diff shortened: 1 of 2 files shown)


rockymadden:
Closed per [thread feedback](https://uipath-product.slack.com/archives/C0AK2KB7KPZ/p1788534281063059). Recording why, for anyone who finds this later.

The orchestrator is the wrong layer. `experiments.yaml` already carries `defaults.agent.system_prompt`, and the override chain is `experiment.variant > task > experiment.defaults`, so this is a config change rather than a code change.

Two things this PR got wrong:

**Scale.** I sized the simulated-task exposure from the `uipath-maestro-flow` suite alone (9 tasks) and assumed the function-boundary split in `_evaluation_loop` covered it. Repo-wide it is **319 simulated tasks** against 973 non-simulated, with `uipath-troubleshoot` alone holding 297.

**Layer.** Even with the simulated path excluded, baking a behavioral policy into the orchestrator makes it unconfigurable per experiment. The config mechanism already existed and I did not look for it.

Replacement work moves to `UiPath/skills`: the headless declaration goes in `tests/experiments/*.yaml`, and the skill-side instruction for what to *do* when headless stays in UiPath/skills#3079.

One open problem for whoever picks up the split: **no tag or path convention distinguishes the 319 simulated tasks**, and they are not confined to a directory (`uipath-maestro-flow` keeps them under `interactive/`, `uipath-troubleshoot` does not). A separate experiment config for them needs either a new tag on 319 files or a selector in `coder-eval run` — the latter would land in this repo.

PR #161, opened 2026-09-09
feat(evaluate): grade a `driver: docker` row inside a container of its own image

akshaylive:
## What

`coder-eval evaluate <run_dir>` and `run --resume` now grade a `driver: docker`
row **inside a container of the task's own image**, instead of on the host.

That is the only place its criteria mean what they meant during the run.
`tasks/byod_smoke_test.yaml` asserts `test -f /opt/byod_marker`, a file baked
into its image: the identical row scores `SUCCESS 1.000` graded in a container
and `FAILURE 0.000` graded on the host — the host answering a question nobody
asked. A container-graded row carries no `graded_on_host` stamp, so it is
indistinguishable from a `run` row, which is the parity that makes the
execute/evaluate split honest.

`--allow-host-grading` survives as the escape hatch (no docker here; criteria
known to be host-portable) and still stamps.

So the resulting model is:

| | `execute` | `evaluate` / `run --resume` |
|---|---|---|
| `driver: tempdir` | host | host |
| `driver: docker` | container | **container** (a second, separate one) |

## Review

The first commit was reviewed across 8 axes and iterated on twice. The review
found 2 score-altering defects, 1 security regression and 4 correctness bugs;
all are fixed in `03f79f4e`. The highest-value ones:

**The dispatch was outside the trust gate.** A run dir is a shareable artifact,
so its recorded config is untrusted — but the dispatch builds a container from
the recorded `sandbox.docker` block. The record names the image, and a container
of it runs with the default credential allowlist forwarded in and a copy of
`~/.claude` mounted. `embedded_commands` walked only `success_criteria` and
`post_run`, so a run dir whose criteria were all `file_exists` reached
`docker run` **with no flags** — where the previous release refused outright.
This is the same blind spot the function's own docstring already describes for
`--copy` provisioning, one layer up. Now scanned on the in-place path, exactly
as `post_run` is.

**Version skew turned a grade into a fresh agent run.** An image predating this
change ignores the unknown `regrade` key and runs the agent; the host folded that
fabricated trajectory back as the recorded row's verdict, and billed for it.
`_assert_regrade_honored` is the exact sibling of `_assert_grade_honored`.

**The record named a path that exists on no host.** A container run recorded
`source_file` as `/work/task_dir/task.yaml`, so the `task_file is None` guard
passed and `_prepare_task_dir_mount` silently mounted nothing — every `$TASK_DIR`
criterion resolving against the wrong tree, with no error.

**The grading container inherited the caller's `run_dir`.** On `run --resume`
that is the executed row's own directory, where a dead grading container's stale
pre-grade `task.json` was read back as a successful grade, and `docker.log` was
truncated.

**The workspace mount skipped `grant_container_access`** — the only
framework-owned mount to do so while the DAC caps are dropped, so a host-owned
workspace failed EACCES and booked a gating `0.0` that reads as an agent failure.

## Known limitation (documented, warned at dispatch)

The grading container is a **second, fresh** container. Only the workspace
crosses, and `pre_run` is not re-run — so a criterion depending on state
`pre_run` put outside the workspace does not see it.
`tasks/samples/skillsbench/3d-scan-calc` symlinks `/root/mass_report.json` in
`pre_run` and its verifier asserts that path, so it would score 0.000 for a
trajectory `run` scores 1.000. Re-running `pre_run` would trade this for the
deliverable-clobbering bug `_skip_pre_run_for_adopted` exists to prevent, so it
is warned loudly rather than silently wrong. Same for `dockerfile_path` image
drift between the two phases — there is no `reference_digest`-style pin on the
image yet.

## Lint

**CE056** — no bare `CODER_EVAL_IN_CONTAINER` literal outside
`models/container_paths.py`. The migration converted all four *readers* and left
the single *writer*, so a rename would have disarmed the reference anti-cheat
window, the reference mount, the grading-container recursion guard and the
watchdog together, with nothing failing. CE052 can't catch it — that rule
inspects `if` guards, and the writer is not one.

## Verification

- `ruff check` + `ruff format --check`: clean
- `pyright`: 0 errors
- `pytest`: **5432 passed, 6 skipped**
- Real-docker end-to-end confirmation of both grading entry points
- New tests for the container half (was 48.73% covered), the dispatch ordering,
  the skew guard, the trust gate, and the fail-closed baseline


uipreliga (commented):
# Review: coder_eval — pr:161 (20 files) axis:1,2,3,4,5,6,7,8

Scope: pr:161 (20 files) axis:1,2,3,4,5,6,7,8 · branch `akshaya/dockerized_eval` · `03f79f4` · 2026-09-09T22:38Z · workflow variant

**Change class:** complex — introduces a new container-dispatch control path for detached grading (`_should_grade_in_container` -> `_grade_in_container`), a new trust gate over recorded config, new `Sandbox.adopt` lifecycle semantics, and two new lint rules; correctness requires reasoning about security boundaries and status-classification chains

Architecture, typing, and test discipline stay strong (8.3/10 overall, 9.5 architecture, 9.7 type safety), but the new container-grading path is the weak axis (6/10): it can change a task's score and final_status for identical agent output — a skipped `pre_run` silently turns a 1.000 row into 0.000, a dropped `recorded_task_file` corrupts the row's recorded source path, and a `dockerfile_path` rebuild grades against an unrecorded image — with none of these pinned by tests, so the bottom line is that the feature is well designed but must not ship until the verdict-changing paths are made durable, stamped, and covered.

## Summary

| 1. Code Quality & Style | 6.7 / 10 | 0 | 2 | 2 | 3 | The container-grading path drops `recorded_task_file`, so a graded docker row re-records the container-only `/work/task_dir/task.yaml` |
| 2. Type Safety | 9.7 / 10 | 0 | 0 | 0 | 3 | `_fold_back_container_grade` couples to DockerRunner's log filename by string literal across packages, and drifts silently |
| 3. Test Health | 9 / 10 | 0 | 1 | 0 | 0 | The new container-dispatch grading seams are unpinned by tests — reverting or mutating each one leaves the full suite green |
| 4. Security | 8.4 / 10 | 0 | 0 | 3 | 1 | Container-dispatch trust gate discloses only `sandbox.docker.*`: the task-dir copy from the recorded `source_file`, the auto-mounted `plugins`/`template_sources`/`system_prompt_file` host paths, and the writable `~/.claude` copy all reach the record-named image undisclosed |
| 5. Architecture & Design | 9.5 / 10 | 0 | 0 | 1 | 0 | A container-graded row emits no `CoderEval.Task.End` telemetry, breaking the "container silent, host emits once" invariant that `batch.py` compensates for on the run path |
| 6. Error Handling & Resilience | 8 / 10 | 0 | 1 | 2 | 0 | A failed grading container deletes its scratch dir with `docker.log` inside it, and the error tells the operator to read that now-deleted path |
| 7. API Surface & Maintainability | 9 / 10 | 0 | 0 | 2 | 0 | `--allow-host-grading` still documented as "instead of refusing" in both USER_GUIDE flag tables (docs/USER_GUIDE.md:45, :273) and the `run` CLI help (cli/run_command.py:221,224) after this PR flipped the `driver: docker` default from refuse to grade-in-container |
| 8. Evaluation Harness Quality | 6 / 10 | 1 | 1 | 0 | 0 | Container grading silently flips a passing row to 0.000 when pre_run seeded state outside the workspace — 3 of 10 in-tree docker tasks, no durable marker (only a console warning at regrade.py:764) |

**Overall Score**: 8.3 / 10 · **Weakest Axis**: Evaluation Harness Quality at 6 / 10
**Totals**: 🔴 1 · 🟠 5 · 🟡 10 · 🔵 7 across 8 axes.

## Blockers

1. **[Axis 1] The container-grading path drops `recorded_task_file`, so a graded docker row re-records the container-only `/work/task_dir/task.yaml`** (`src/coder_eval/cli/run_task_internal_command.py:282`) — The PR adds `Orchestrator(recorded_task_file=...)` specifically so a container run stops recording `/work/task_dir/task.yaml` as `task_config.source_file` (orchestrator.py:448-455: "Recording it made a container row's `source_file` name a path that exists on no host"). `run_task_internal_command` computes that value at line 235 (`recorded_task_file = Path(host_task_file_raw) if host_task_file_raw else None`) and forwards it on the ordinary branch at line 303 (`recorded_task_file=recorded_task_file`) — but the new regrade branch at lines 281-292 does not:

```python
    if regrade_raw:
        _grade_recorded_run(
            task=task,
            authored_task=authored_task,
            input_dir=input_dir,
            output_dir=output_dir,
            runtime_task_file=runtime_task_file,
            source_yaml=source_yaml,
            variant_id=variant_id,
            replicate_index=replicate_index,
        )
        return
```

`_grade_recorded_run` has no such parameter, and neither does `regrade_in_place` (`src/coder_eval/orchestration/regrade.py:831-843`), whose `Orchestrator(...)` call at regrade.py:898-910 omits it entirely — so `self.recorded_task_file` falls back to `task_file`, which here is `runtime_task_file = task_dir / "task.yaml"` = `/work/task_dir/task.yaml` (run_task_internal_command.py:253). `Orchestrator._finalize_result` then rewrites the record unconditionally at orchestrator.py:1206-1211 (`source_file=str(self.recorded_task_file) if self.recorded_task_file else None`), and `_fold_back_container_grade` copies those bytes verbatim into the host row (regrade.py:665-666). Net effect: every container-graded row's `task.json` regresses `source_file` back to a path that exists on no host — the exact defect `recorded_task_file` was added for — and a subsequent `evaluate <run_dir>` on that row now trips `_grade_in_container`'s own "a task file that is not on this host" refusal (regrade.py:731-737). Fix: add a `recorded_task_file: Path | None = None` parameter to `regrade_in_place` (passed straight through to `Orchestrator`) and to `_grade_recorded_run`, and forward the already-computed value at line 282. Add an assertion to `TestDockerRunnerGradingWiring` that a container-graded row's `task_config.source_file` equals the host path, mirroring `test_a_grading_run_forwards_the_hosts_own_task_file_for_the_record` (tests/test_regrade.py:655-679), which today only checks the wire format and not the branch that consumes it.
2. **[Axis 1] `task_from_prior` gains a `grade_in_place` flag that is the exact negation of the `include_setup_phase` flag beside it — one fact, two parameters, no consistency check** (`src/coder_eval/orchestration/regrade.py:69`) — `task_from_prior` now takes both `include_setup_phase: bool = True` (line 68) and `grade_in_place: bool = False` (line 69), and the sole production caller passes them as literal complements (`src/coder_eval/cli/evaluate_command.py:167-179`):

```python
            setup_will_run = not resolve_grade_in_place(target, in_place)
            task, source_yaml = task_from_prior(
                ...
                include_setup_phase=setup_will_run,
                grade_in_place=not setup_will_run,
```

The two are definitionally the same lever: the docstring at regrade.py:153-158 defines `include_setup_phase` as "the two capability families that exist only on the `--copy` path", and `grade_in_place` is "not the `--copy` path". `_fall_back_to_source` (regrade.py:330-331) then carries both through as well. This is a second source of truth for one boolean, with no validator rejecting the two incoherent pairings (`include_setup_phase=True, grade_in_place=True`), and the failure mode is not cosmetic: the flag gates a *security* disclosure — a future caller that passes `include_setup_phase=False` and forgets `grade_in_place=True` silently drops the container-dispatch half of the untrusted-config gate, and nothing fails. Collapse to a single parameter (`grade_in_place: bool = False`, with `include_setup_phase = not grade_in_place` derived internally, or the reverse), which also removes the duplicated derivation expression that now appears verbatim at two sites — regrade.py:113-114 and regrade.py:355-356, both `include_container_dispatch=grade_in_place and _should_grade_in_container(task, allow_host_grading=allow_host_grading)`. Also note that neither `grade_in_place` nor `allow_host_grading` is described in the `task_from_prior` docstring (lines 72-85), which documents only `allow_recorded_commands`.
3. **[Axis 3] The new container-dispatch grading seams are unpinned by tests — reverting or mutating each one leaves the full suite green** (`src/coder_eval/cli/run_task_internal_command.py:377`) — Coverage `term-missing` reports `377-401` for this file, and those 25 lines are all new in this PR: the mount guard (`workspace = Path(CONTAINER_GRADE_WORKSPACE)` / `if not workspace.is_dir(): typer.echo(f"FATAL: the graded workspace was not mounted at {workspace}", err=True)`, lines 377-380), the `asyncio.run(regrade_in_place(... recorded_task=authored_task))` call (382-395), and the `except RegradeError` → exit 2 wrap (396-401). The three new tests in `tests/test_detached_grading_boundaries.py:256-285` only reach the two early returns above them (`'regrade' must be a boolean`, `prior.json is missing`, `not a readable EvaluationResult`). Verified by mutation: deleting the line `                recorded_task=authored_task,` (line 393) AND replacing line 235 `    recorded_task_file = Path(host_task_file_raw) if host_task_file_raw else None` with `    recorded_task_file = None` leaves the full suite at `5430 passed, 8 skipped`. Per the function's own docstring (line 355: "``authored_task`` is what gets RECORDED, so the row keeps saying `driver: docker`") and CLAUDE.md, losing that seam makes the container-graded row record `driver: tempdir`, which then lets a later `evaluate <run_dir>` skip BOTH the host-grading refusal and the `graded_on_host` stamp. Fix: drive `_grade_recorded_run` end-to-end with `regrade_in_place` monkeypatched, asserting (a) it is called with `recorded_task` equal to the authored `driver: docker` task and `workspace == Path(CONTAINER_GRADE_WORKSPACE)`, and (b) an unmounted workspace exits 2 with "was not mounted".
4. **[Axis 6] A failed grading container deletes its scratch dir with `docker.log` inside it, and the error tells the operator to read that now-deleted path** (`src/coder_eval/orchestration/regrade.py:812`) — `_grade_in_container` runs the whole grading container inside `with tempfile.TemporaryDirectory(prefix="coder-eval-grade-") as scratch:` (line 792) and folds artifacts out only on the SUCCESS path — `_fold_back_container_grade(container_run_dir, run_dir)` is at line 827, *after* the `try`. Every failure path exits the `with` block and deletes the scratch dir:

```python
        except (DockerRunError, OSError) as e:
            raise RegradeError(
                f"Grading {task.task_id!r} in a container failed: {e}. Re-run with --allow-host-grading "
```

What is destroyed is precisely the evidence. `DockerRunner.run()` writes `log_path = self.rt.run_dir / "docker.log"` (docker_runner.py:709) into the scratch dir, and `_stream_container_output` sends container stdout+stderr ONLY there — `_grade_in_container` passes no `stream_callback` and no `verbose`, so nothing is echoed. The wrapped message the operator sees is built at docker_runner.py:917-920:

```python
            error = DockerRunError(
                f"Container exited with code {returncode} without producing task.json. "
                + f"See {log_path} for container output."
            )
```

i.e. it instructs the operator to read `/var/folders/.../coder-eval-grade-XXXXXX/docker.log`, a path that no longer exists by the time the message is printed. Two sibling artifacts vanish with it: `_record_build_failure` (docker_runner.py:1064) writes a failed image build's captured build log to `run_dir/docker.log` + a synthetic `BUILD_FAILED` task.json — both into the scratch dir — and `_write_synthetic_task_json` (docker_runner.py:921) does the same for the ERROR case. And the container's own `_grade_recorded_run` FATAL lines (`FATAL: {prior_path} is not a readable EvaluationResult`, run_task_internal_command.py:370; `FATAL: the graded workspace was not mounted at ...`, :375; `FATAL: {e}` for a `RegradeError`, :397) are stderr, merged into stdout by `stderr=asyncio.subprocess.STDOUT` (docker_runner.py:1047-ish `create_subprocess_exec`) and therefore land in that same deleted `docker.log`. Net: an image-build failure, an OOM/exit-137, or any container-side guard produces a message with no reachable log anywhere, on the exact path this feature makes routine for nightly `run --resume` over `driver: docker` rows.

Fix: preserve the diagnostic before the scratch dir goes away. Wrap the `DockerRunner(...).run()` call so the `except` copies `container_run_dir/"docker.log"` to `run_dir/"grade.docker.log"` (the same best-effort copy `_fold_back_container_grade` already does on success, regrade.py:675-681) BEFORE raising, and re-point the `RegradeError` text at that surviving path rather than letting the inner `See {log_path}` sentence through unedited. A `finally:` around the call that always folds out `docker.log` is the smallest change and makes success and failure symmetric; the note in `_fold_back_container_grade`'s docstring that everything but `task.json` "stays in the scratch dir and is discarded with it, which is the point" is right for the success path and exactly backwards for the failure path.
5. **[Axis 8] Container grading silently flips a passing row to 0.000 when pre_run seeded state outside the workspace — 3 of 10 in-tree docker tasks, no durable marker (only a console warning at regrade.py:764)** (`src/coder_eval/orchestration/regrade.py:763`) — `_grade_in_container` dispatches a SECOND, fresh container in which `pre_run` is not re-run (`Sandbox.adopt` sets `was_adopted`), and the only control is a console log at regrade.py:763: `logger.warning("Task %r declares %d pre_run command(s). They ran in the container that executed the agent and are NOT re-run here...")`. That warning is not durable. The codebase already states why, 140 lines earlier, in `stamp_host_grading`'s own docstring (regrade.py:624-627): "Written onto the result, not just logged: a console warning does not travel with ``task.json`` into ``run.json``, the reports or the evalboard." The same argument applies verbatim here and was not followed.

This is not hypothetical and it is not rare. Three of the ten in-tree `driver: docker` tasks match the pattern exactly — `tasks/samples/skillsbench/3d-scan-calc/3d-scan-calc.yaml:52` (`- command: ln -sfn "$PWD/mass_report.json" /root/mass_report.json`), `court-form-filling.yaml:40`, `dialogue-parser.yaml:51` — and the 3d-scan-calc verifier's FIRST assertion is `tasks/samples/skillsbench/3d-scan-calc/verifier/test_outputs.py:11`: `assert os.path.exists("/root/mass_report.json"), "Output file not found"`. In a fresh grading container `/root` is pristine, `test.sh` writes `echo 0 > /logs/verifier/reward.txt`, and the single `score_from_stdout` criterion scores 0.000 → FAILURE for a trajectory `coder-eval run` scores 1.000.

Reachable with no flags at all: the `--allow-recorded-commands` gate only guards `evaluate <run_dir>`; `coder-eval execute <task.yaml>` followed by `coder-eval run --resume <run_dir>` routes the NOT_GRADED row to `_grade_resumed_tasks` (cli/run_command.py:783), which builds the task from the operator's own YAML and calls `regrade_in_place` with no gate. Before this PR that same command was a hard refusal (`grading_sandbox_config` raised), so the PR converts a loud refusal into a silently-wrong verdict. And by design (CLAUDE.md: "A container-graded row carries NO `graded_on_host` stamp, so it is indistinguishable from a `run` row") nothing downstream — run.json, the reports, or the separate eval-runner repo — can tell the difference.

Fix: make it durable and detectable, not merely audible. Write the fact onto the result the way `stamp_host_grading` does — e.g. `result.environment_info["graded_without_pre_run"] = len(task.pre_run)` in `_grade_in_container` — so run.json, the HTML report and the evalboard can filter such rows out of any comparison. Better still, REFUSE by default when `task.pre_run` is non-empty and require an explicit opt-in (the pattern `--allow-host-grading` already establishes for the other unequal-grading route), since a 0.000 that reads as an agent failure is the CE039 shape this feature exists to end.
6. **[Axis 8] A `dockerfile_path` row is graded against a freshly rebuilt image with no image identity recorded, and the dispatch warning USER_GUIDE.md:231 promises does not exist** (`docs/USER_GUIDE.md:231`) — docs/USER_GUIDE.md:229-231 tells operators: "for a `dockerfile_path` task the grading phase re-runs `docker build`, so a Dockerfile or base image that changed between the two phases yields a different grading image. Both cases are warned about at dispatch". The second sentence is false. `_grade_in_container` (regrade.py:740-772) emits exactly two log records — `logger.info("Grading %r in a container of its own image...")` at 740 and the `pre_run` `logger.warning` at 763 — and `grep -n "dockerfile_path" src/coder_eval/orchestration/regrade.py` returns only lines 255-261, inside `embedded_commands`' trust-gate inventory. There is no dispatch-time warning for the rebuild case anywhere.

The underlying gap is a scoring one, which is why this is not merely a docs nit. `DockerRunner._resolve_image` (docker_runner.py:1319-1338) unconditionally re-runs `docker build -t coder-eval-task-<id>:built -f <dockerfile> <context>` on every dispatch, and the deterministic tag means the grading image REPLACES the run's image under the same name. Nothing pins or records the image identity on either side: `grep -rn 'environment_info\[' src/coder_eval/orchestrator.py src/coder_eval/isolation/docker_runner.py` shows no image, tag or digest key. So for a task whose Dockerfile, build context, or `FROM coder-eval-agent:<version>` base moved between `execute` and `evaluate`, the criteria run against a different filesystem than the agent did, and the resulting `score` / `final_status` change for identical agent output with no warning, no stamp, and no way for the eval-runner consumer to detect it.

Fix, in order of value: (1) record the resolved image reference (and, cheaply, `docker image inspect -f '{{.Id}}'`) into `environment_info` on both the run and the grade, and refuse — or at minimum warn loudly at dispatch — when they differ; (2) until (1) lands, add the warning the guide already promises, next to the `pre_run` one at regrade.py:763, gated on `task.sandbox.docker.dockerfile_path`; (3) if neither is done in this PR, correct USER_GUIDE.md:231 so it does not describe a control that is absent.

1. **[Axis 1] `embedded_commands`' hand-mirrored docker-argv inventory renders and counts continuation fragments as separate shell commands** (`src/coder_eval/orchestration/regrade.py:254`) — The function's contract is stated in its own first docstring line (regrade.py:151): "Every shell command a rebuilt task definition would run on this host." The new block at lines 254-273 appends items that are not commands — two-space-indented argv fragments and a prose annotation:

```python
            commands.append(f"docker build -f {docker.dockerfile_path}")
            for key, value in docker.build.args.items():
                commands.append(f"  --build-arg {key}={value}")
...
            commands.append(f"docker run {docker.image} (with your credentials in its environment)")
        for mount in docker.extra_mounts or []:
            commands.append(f"  -v {mount}")
```

`check_embedded_commands` then does `rendered = "; ".join(commands)` (line 307) and interpolates `len(commands)` into the operator-facing refusal (line 310). Verified by running the real code against a `dockerfile_path` task with two build args and one extra mount:

```
The config recorded in /run/task.json would run 4 shell command(s) on this host with your environment: docker build -f Dockerfile;   --build-arg FOO=bar;   --build-arg BAZ=qux;   -v /a:/b
```

That is one docker invocation reported as four shell commands, rendered as a semicolon-separated list in which `  --build-arg FOO=bar` reads as a standalone command. The consent prompt is the one place this text has to be exact. Same root cause raised complexity from C(17) on `origin/main` to D(25) at HEAD (radon, verified against a base copy of the file). Fix both together: extract a `_container_dispatch_commands(docker) -> list[str]` helper that emits ONE string per dispatch (e.g. `docker build -f Dockerfile --build-arg FOO=bar --build-arg BAZ=qux -v /a:/b`), restoring both the documented contract and the count. Note the only tests for this block (tests/test_regrade.py:749-786) exercise the `image` branch, never `dockerfile_path`, which is why the fragment rendering was never seen.
2. **[Axis 1] `_assert_regrade_honored` repeats `_assert_grade_honored`'s 7-line quarantine block verbatim except for the sidecar suffix and log wording** (`src/coder_eval/isolation/docker_runner.py:968`) — docker_runner.py:968-977 and docker_runner.py:1018-1026 are the same seven lines differing only in the suffix string and the log wording:

```python
        if task_json is not None:
            sidecar = task_json.with_suffix(task_json.suffix + ".rerun")
            try:
                os.replace(task_json, sidecar)
                logger.warning("Quarantined the refused re-run record to %s", sidecar)
            except OSError as exc:
                logger.warning("Could not quarantine %s: %s", task_json, exc)
```

versus the `.graded` variant at 1018-1026; `_handle_malformed_task_json` (lines 1050-1058) is a third instance of the same move-aside-best-effort pattern with a `.malformed` suffix. The new copy's own comment concedes the duplication ("Same sidecar pattern as `_assert_grade_honored`"). Extract `def _quarantine_record(self, task_json: Path | None, suffix: str, label: str) -> None:` and call it from both assert methods, so a future change to the quarantine semantics (e.g. not clobbering an existing sidecar) cannot land on one refusal path and miss the other.
3. **[Axis 4] Container-dispatch trust gate discloses only `sandbox.docker.*`: the task-dir copy from the recorded `source_file`, the auto-mounted `plugins`/`template_sources`/`system_prompt_file` host paths, and the writable `~/.claude` copy all reach the record-named image undisclosed** (`src/coder_eval/orchestration/regrade.py:269`) — `embedded_commands`' new `include_container_dispatch` branch (regrade.py:254-273) enumerates only the docker block — the whole operator-facing disclosure of the container is the single line at 269: `commands.append(f"docker run {docker.image} (with your credentials in its environment)")` plus `f"  -v {mount}"` for `docker.extra_mounts` (271). But the dispatch mounts three more host locations that the same untrusted record chooses, and none of them is named:

1. `_grade_in_container` accepts any recorded task file that merely EXISTS — `regrade.py:730`: `if task_file is None or not task_file.is_file():` — and `task_file` on the detached path is `Path(prior.task_config.source_file)` (evaluate_command.py:183). DockerRunner then does `source = self.rt.task_file.parent.resolve()` (docker_runner.py:1241) and `shutil.copytree(source, task_dir_copy, ...)` (docker_runner.py:1245), mounting the copy at `/work/task_dir`. A record whose `source_file` is `~/.ssh/config` copies all of `~/.ssh` into the image the same record names. The function's own docstring concedes this and defers to the gate — but the gate never says it.
2. `_build_argv` auto-mounts every `agent.plugins[].path` (docker_runner.py:1692) and every `TemplateDirSource.path` (docker_runner.py:1700) read-only at its host path via `argv.extend(["-v", f"{target}:{target}:ro"])` (docker_runner.py:1689). `_sensitive_source_paths` only *logs a warning* for a fixed list (`~/.ssh`, `~/.aws`, `~/.gnupg`, `~/.config/gh`, `~/.kube`, `/etc`) — and this file's own gate docstring states the governing principle: "A warning is not a control: it is printed as the command is already being prepared."
3. The `~/.claude` copy (including `.credentials.json`, which is absent from `CLAUDE_COPY_IGNORE`, docker_runner.py:117-136) is mounted read-WRITE; the docstring at regrade.py:194 names it, the operator-facing string at 269 says only "in its environment".

Network defaults to `--network bridge` unless the record says otherwise, so the container can exfiltrate all of it. Fix: (a) render the resolved task-dir source and every `_auto_mount` target into the `include_container_dispatch` command list so the operator approves what is actually exposed; (b) containment-check the recorded `source_file` against the run directory (or the operator's task tree) and refuse an outside path unless `evaluate <task.yaml> <run_dir>` is used — existence is not a trust property, exactly as the docstring at regrade.py:748-753 already argues. CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:N/A:N
4. **[Axis 4] Container grading recursively chmods the graded workspace world-writable on the host and never restores it — new, permanent exposure for an operator-supplied `--workspace` outside the run dir** (`src/coder_eval/isolation/docker_runner.py:694`) — New in this PR: `await asyncio.to_thread(grant_container_access, self.grade_workspace, writable=True)` (docker_runner.py:694). `grant_container_access` is `chmod -R o+rwX` semantics (`extra = 0o006 if writable else 0o004`, docker_runner.py:521-536) and there is no restore anywhere — unlike `input_dir`/`output_dir`, which are harness-created, disposable staging. `grade_workspace` is not: it is `default_workspace(run_dir, prior)` (which can be the run's preserved `artifacts/` tree) or, with `--workspace`, an arbitrary operator directory — the option's own help says "Grade this directory instead of the run's own artifacts ... (e.g. a verifier's /app)" (evaluate_command.py:255-260), and the comment at docker_runner.py:689 explicitly anticipates "an operator-supplied `--workspace`". After one `coder-eval evaluate`, every file and directory in that tree is world-writable on the grading host, permanently. On a shared/CI host any other local uid can then rewrite the artifacts a criterion reads, i.e. change the verdict, or plant executables in the tree. Fix: snapshot the pre-widening modes and restore them in the same `finally` that removes `staging`, or widen only the specific paths the container must write and refuse to widen a `--workspace` that is not inside the run directory without an explicit opt-in. CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:N/I:H/A:N
5. **[Axis 4] `_fold_back_container_grade` copies the container log with `shutil.copy2` (regrade.py:681), which follows a destination symlink — the exact primitive the `write_text_atomic` call seven lines above (line 674) exists to prevent, and which three sibling sites in the same PR guard with `is_symlink()`** (`src/coder_eval/orchestration/regrade.py:681`) — Line 674 of the same function writes the verdict safely — `write_text_atomic(run_dir / TASK_JSON_FILENAME, graded.read_text(encoding="utf-8"))` — precisely because, per `path_utils.py:63-68`, "a pre-planted `task.json.tmp` *symlink* in a shared run directory makes this an arbitrary-file-overwrite primitive." Seven lines later the sibling write has no such guard: `shutil.copy2(container_log, run_dir / "grade.docker.log")` (regrade.py:681). `shutil.copy2` opens the destination with `open(dst, 'wb')`, which follows a symlink at `dst`, so a pre-planted `grade.docker.log` symlink in the destination directory redirects the container's log content to any path the invoking uid can write. It is reachable wherever `run_dir` is a directory the grader did not create: `run --resume` passes the executed row's own directory (`run_command.py:783-793`, `run_dir=rt.run_dir`), and `evaluate --run-dir <shared run dir>` targets an artifact directory the operator obtained from elsewhere. The surrounding `contextlib.suppress(OSError)` also means the redirect leaves no trace. Fix: refuse when the destination `is_symlink()`, or write through `os.open(dest, O_CREAT|O_EXCL|O_NOFOLLOW)` the way `write_text_atomic` does. CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:N/I:H/A:N
6. **[Axis 5] A container-graded row emits no `CoderEval.Task.End` telemetry, breaking the "container silent, host emits once" invariant that `batch.py` compensates for on the run path** (`src/coder_eval/orchestration/regrade.py:828`) — `_grade_in_container` ends with `_fold_back_container_grade(container_run_dir, run_dir)` / `return result` (lines 827-828) and never emits a task-usage event. Every sibling path does:

- Host grading (`--allow-host-grading`, or any `tempdir` row) goes through `regrade_in_place` -> `Orchestrator.run()` -> `_finalize_result`, which at `orchestrator.py:1235-1236` does `name, props = build_task_event(...)` / `track_event(name, props)` unconditionally.
- The docker **run** path is compensated host-side on purpose: `orchestration/batch.py:173-177` re-emits after parsing the container result, with the comment "The in-container `_finalize_result` can't emit task telemetry ... so emit the Task.End event host-side here — keeping docker runs at parity with the in-process path."
- `DockerRunner._build_argv` hard-disables telemetry in every container it starts: `argv += ["--env", "TELEMETRY_ENABLED=false"]` (docker_runner.py:1598), whose comment at line 1595 states the invariant verbatim: `The invariant is "container silent, host emits once"`.

The grading container inherits the "container silent" half (it is started by the same `_build_argv`) but nothing supplies the "host emits once" half, so the event is dropped entirely. Result: after this PR, `coder-eval evaluate <run_dir>` and `coder-eval run --resume` over a `driver: docker` row publish a verdict that never reaches usage telemetry, while the byte-identical operation on a `tempdir` row — or the same docker row with `--allow-host-grading` — does. That is a silent, driver-dependent hole in the exact metric the split was meant to keep at parity with `run`.

Fix: mirror `batch.py:173-177` at `regrade.py:828` — `from coder_eval.orchestrator import build_task_event; from coder_eval.telemetry import track_event; name, props = build_task_event(result, driver="docker", variant_id=variant_id); track_event(name, props)` — or, better, hoist that host-side compensation into a single helper both `batch.py` and `_grade_in_container` call, so the next container-dispatching caller cannot forget it.
7. **[Axis 6] Container-graded rows' log artifacts diverge from the run-layout contract: no `grade.log`, and an undocumented `grade.docker.log`** (`src/coder_eval/orchestration/regrade.py:681`) — `_fold_back_container_grade` copies exactly two things out of the grading container's scratch dir — `task.json` (line 674) and `docker.log` renamed to `grade.docker.log`:

```python
    container_log = container_run_dir / "docker.log"
    if container_log.is_file():
        with contextlib.suppress(OSError):
            shutil.copy2(container_log, run_dir / "grade.docker.log")
```

But the grading pass's own log is a THIRD file. Inside the container, `regrade_in_place` builds an `Orchestrator` with `prior_result=prior`, and `orchestrator.py:645` selects the log file as `task_log_path(self.run_dir, regrade=self.prior_result is not None)` — which `path_utils.py:184` resolves to `run_dir / GRADE_LOG_FILENAME` (`"grade.log"`). The in-container `run_dir` is `CONTAINER_OUTPUT_DIR`, bind-mounted from the scratch dir, so `grade.log` is written there and deleted with the scratch dir. Nothing folds it back.

That log is a documented part of the run-directory contract — `.claude/shared/run-layout.md:17`: "`grade.log` — present only after a DETACHED grade over this directory (`coder-eval run --resume`). The grading pass's own log." It holds the per-criterion evaluation detail (`run_command` stdout/stderr, `llm_judge` prompt/verdict, checker warnings), which is the only durable record of WHY a criterion scored what it scored. So the identical detached grade produces `grade.log` for a `driver: tempdir` row and no `grade.log` for a `driver: docker` row — a driver-dependent hole in a documented artifact, on the exact rows whose criteria are hardest to reason about from the host.

Fix: copy `container_run_dir / GRADE_LOG_FILENAME` to `run_dir / GRADE_LOG_FILENAME` in `_fold_back_container_grade` alongside the `docker.log` copy, using `GRADE_LOG_FILENAME` from `path_utils` rather than a literal. Unlike `docker.log`, the name does NOT collide on the `run --resume` path (a `run --resume` grade is the only thing that writes `grade.log` into that directory), so no phase-specific rename is needed.
8. **[Axis 6] `_fold_back_container_grade`'s IO escapes as a raw `OSError` on the `evaluate <run_dir>` path, losing an already-computed verdict (regrade.py:673-674, called at :827 outside the try that closes at :823)** (`src/coder_eval/orchestration/regrade.py:674`) — The mandatory half of the fold-back is unguarded:

```python
    run_dir.mkdir(parents=True, exist_ok=True)
    write_text_atomic(run_dir / TASK_JSON_FILENAME, graded.read_text(encoding="utf-8"))
```

All three calls (`mkdir`, `read_text`, `write_text_atomic`) can raise `OSError`, and this line sits OUTSIDE the `try` in `_grade_in_container` (the `try` closes at line 823; the fold-back is at line 827). So the exception leaves `orchestration/` as a raw `OSError`, not the `RegradeError` the layer contract promises. The two callers then disagree, and both outcomes are wrong in a different direction:

* `evaluate <run_dir>`: `run_evaluation` guards `asyncio.run(_setup_and_run())` with `except RegradeError as e:` only (evaluate_command.py:472). An `OSError` sails past it into Typer as a stack trace — after a grade that already SUCCEEDED and whose verdict is in the scratch dir about to be deleted.
* `run --resume`: `_grade_resumed_tasks` catches `(RegradeError, OSError, RuntimeError, ValueError)` (cli/run_command.py:790), so the row is folded back with its ORIGINAL ungraded result and an `error_message` reading "Grading failed during --resume" — reporting a computed, correct verdict as a grading failure and discarding it.

This also contradicts the convention stated 190 lines below in the same file's sibling writer, `_write_back` (evaluate_command.py:619-625), whose `except OSError` comment reads: "Never fail the grade over the write-back: the verdict was computed and already printed, and the fresh run dir holds its own task.json." `_fold_back_container_grade`'s docstring asserts the opposite ("mandatory on the record: a grade that cannot write its verdict is a failure") — that is a defensible position, but then it must fail as a `RegradeError` naming the destination and the next step, not as an unwrapped `OSError` whose handling differs per caller. Fix: wrap the `mkdir`/`read_text`/`write_text_atomic` trio in `try/except OSError` inside `_fold_back_container_grade` and `raise RegradeError(f"Graded {task_id!r} but could not write the verdict to {run_dir / TASK_JSON_FILENAME}: {e}") from e`, so both callers get the same, actionable outcome.
9. **[Axis 7] `--allow-host-grading` still documented as "instead of refusing" in both USER_GUIDE flag tables (docs/USER_GUIDE.md:45, :273) and the `run` CLI help (cli/run_command.py:221,224) after this PR flipped the `driver: docker` default from refuse to grade-in-container** (`docs/USER_GUIDE.md:45`) — This PR changed the default for a `driver: docker` row from *refuse* to *grade in a container of the task's own image*, but the two flag tables — the primary reference surface for the flag — still describe the old default. `docs/USER_GUIDE.md:45` (the `run` table): "`--allow-host-grading` | `--resume` only. Grade an executed-but-ungraded `driver: docker` row on this host **instead of refusing**; the row is stamped `graded_on_host`." `docs/USER_GUIDE.md:273` (the `evaluate` table): "Grade a `driver: docker` task on this host **instead of refusing**." Neither is true any more: `_should_grade_in_container` (`src/coder_eval/orchestration/regrade.py:651`) returns True whenever the driver is docker, the flag is absent and `IN_CONTAINER_ENV != "1"`, and `regrade_in_place` then dispatches `_grade_in_container` at line 875 — on BOTH the `evaluate` and the `run --resume` paths (`cli/run_command.py:783` forwards `allow_host_grading` into the same call). The prose 50 lines above (`docs/USER_GUIDE.md:212-222`) states the new behaviour correctly, so the file contradicts itself. The identical stale sentence also survives in the CLI `--help` at `src/coder_eval/cli/run_command.py:221` — "Grading cannot start a container, so such criteria run against a filesystem lacking the container's paths and toolchain... Without this they are refused and stay ungraded" — which is wrong twice over (grading now can and does start a container, and without the flag the row is graded in one rather than left ungraded). Rewrite all three to "grade on this host instead of in a container of the task's own image (the default)", matching the wording already used in `cli/evaluate_command.py:303-308`.
10. **[Axis 7] Docker refusal message directs the operator to remove a `--copy` flag that the two-argument form does not have** (`src/coder_eval/orchestration/regrade.py:609`) — `grading_sandbox_config` now ends its refusal with: `"Re-run WITHOUT --copy to grade it in a container (the default for a run directory), " + "or with --allow-host-grading to accept host grading anyway."` (lines 609-610). By the function's own docstring at lines 573-575, exactly two shapes reach this branch: "the ``--copy`` path, which cannot adopt a container workspace, **and the two-argument ``evaluate <task.yaml> <dir>`` form**." For that second shape the first half of the advice is inert: no `--copy` was passed (copy is the default for a plain work directory per `resolve_grade_in_place`, `cli/evaluate_command.py:68`), and passing `--in-place` does not help either, because the container route is gated on `delegates_to_regrade = grade_in_place and prior is not None` (`cli/evaluate_command.py:402`) and `prior` is `None` in the two-argument form. So `coder-eval evaluate tasks/byod_smoke_test.yaml ./dir` prints an instruction that cannot work, and the only real way forward — `--allow-host-grading` — is presented as the fallback. The pre-PR message (`"Re-run with --allow-host-grading to accept that, or grade on a machine that reproduces the container."`) was applicable in both shapes; this change made it worse for one of them. Either make the `--copy` sentence conditional on the caller actually being in run-directory mode, or restore shape-neutral wording such as "Grade a run directory in place (the default) to get a container, or pass --allow-host-grading to grade here."

## Nits

1. **[Axis 1] A pre-existing comment block was split from the code it documents by the newly inserted `host_task_file` stanza** (`src/coder_eval/cli/run_task_internal_command.py:227`) — Lines 227-237 now read as one nine-line comment followed by two unrelated assignments:

```python
    # Docker WORKDIR alignment: the host resolves the concrete WORKDIR
    # (config value / "auto" -> `docker inspect` / fallback) and forwards it here.
    # Absent -> None -> standard run_dir/artifacts workspace.
    # What task.json RECORDS as the task's source path, as distinct from the
    # path this process resolves TASK_DIR against ...
    host_task_file_raw = context.get("host_task_file")
    recorded_task_file = Path(host_task_file_raw) if host_task_file_raw else None
    workspace_dir_raw = context.get("workspace_dir")
```

The first three lines describe `workspace_dir_raw` (line 236), not `host_task_file_raw` (line 234) — "Absent -> None -> standard run_dir/artifacts workspace" now sits directly above the task-file lines and reads as describing them. Move the `host_task_file` stanza below the `workspace_dir` pair, or vice versa, so each comment abuts its own code. While there: line 223 introduces `regrade_raw` but, unlike its sibling `grade` (line 217, `grade: bool = grade_raw`), never binds a typed `regrade: bool`, so the consumer at line 281 reads `if regrade_raw:` — mirror the sibling for symmetry.
2. **[Axis 1] USER_GUIDE.md:220 restates the `--allow-host-grading` escape hatch near-verbatim in one section (plus a mis-pointed pronoun)** (`docs/USER_GUIDE.md:220`) — Lines 220-222 say: "`--allow-host-grading` is the escape hatch: no docker here, or criteria you know are host-portable. It grades on this machine instead, and stamps the row `graded_on_host` so it is never silently compared with a container-graded one." Lines 243-246 then repeat it: "`--allow-host-grading` keeps its meaning as the escape hatch: grade here anyway, for a machine with no docker or for criteria you know are host-portable. It still stamps `graded_on_host` in `environment_info`, so such a row is never silently compared with a container-graded one." Line 240-241 likewise re-states "A container-graded row carries no `graded_on_host` stamp, exactly like a row `coder-eval run` produced", already covered by the first paragraph. Delete the second paragraph (243-246) and the trailing clause of 240-241. Separately, line 224 opens "Two limits are worth knowing before you rely on **it**" immediately after the `--allow-host-grading` paragraph, so "it" reads as that flag — but the two limits described (fresh container, `pre_run` not re-run, `docker build` re-run) are properties of *container* grading, the opposite branch. Rewrite as "before you rely on container grading".
3. **[Axis 1] Redundant local `import json` shadowing the module-level import in new/edited tests** (`tests/test_regrade.py:667`) — `tests/test_regrade.py` imports `json` at module level (line 13), then `test_a_grading_run_forwards_the_hosts_own_task_file_for_the_record` shadows it with a local `import json` at line 667 (alongside a local `import asyncio` at 666, even though the module runs under `asyncio_mode=auto` and the sibling tests in the same class are plain `async def`). Drop both local imports and make the test `async def`, matching `test_a_grading_run_stages_the_prior_row_and_flags_the_regrade` (line 593). Separately, the same six-line `task_file.write_text(...)` + `ResolvedTask(task=_docker_task(), ...)` construction is repeated in three places — `TestDockerRunnerGradingWiring._runner` (line 573), `TestRegradeSkewGuard._runner` (line 688), and inline in `test_an_ordinary_run_is_never_checked` (lines 728-735) — where a single module-level `_docker_runner(tmp_path, **kw)` helper would serve all three.
4. **[Axis 2] `_fold_back_container_grade` couples to DockerRunner's log filename by string literal across packages, and drifts silently** (`src/coder_eval/orchestration/regrade.py:675`) — regrade.py:675 reads `container_log = container_run_dir / "docker.log"` and line 681 writes `shutil.copy2(container_log, run_dir / "grade.docker.log")`, while the producer of that file is `docker_runner.py:708` / `:1077` (`log_path = self.rt.run_dir / "docker.log"`) — three unrelated string literals in two packages. Because the copy is guarded by `if container_log.is_file():`, a rename on the docker_runner side degrades the fold-back to a silent no-op: the grading container's log is discarded with the scratch tempdir and nothing reports it. This is precisely the shape the repo codified as CE053 (`TASK_JSON_FILENAME`) and again as CE056 (`IN_CONTAINER_ENV`) in this same PR. Promote the name to a constant in `path_utils.py` next to `PRIOR_RESULT_FILENAME` (e.g. `DOCKER_LOG_FILENAME` / `GRADE_DOCKER_LOG_FILENAME`), import it in both modules, and consider widening CE053's `_run_record_filenames()` to cover it so the constant does not become the second source of truth its own docstring warns about.
5. **[Axis 2] New DockerRunner test doubles accept `**kw` with no `spec`, so a constructor-signature drift on the grading seam cannot fail the suite (and pyright does not check tests)** (`tests/test_regrade.py:415`) — `class _FakeRunner: def __init__(self, rt, **kw)` at tests/test_regrade.py:415 and :473, and `class _Boom: def __init__(self, rt, **kw)` at :538, are substituted for the real class via `monkeypatch.setattr(dr, "DockerRunner", _FakeRunner)`. Every parameter is untyped and `**kw` swallows anything, so if `_grade_in_container` (regrade.py:812-822) later renames or adds a required kwarg — `prior_result=` / `grade_workspace=` / `preservation_mode=` — the doubles keep accepting it and only the one explicit `captured["prior_result"]` assertion at :450 would notice. The four in-scope test files are outside pyright's `include` (`coder_eval/` only), so nothing else catches it either. Per shared-rubric criterion 17, bind the double to the real shape: `Mock(spec=DockerRunner)`, or annotate the double's `__init__` with the real signature (`rt: ResolvedTask, *, preservation_mode: PreservationMode, prior_result: EvaluationResult, grade_workspace: Path`) so a production rename becomes a TypeError in the test rather than a silent pass.
6. **[Axis 2] `sandbox` widened to `Sandbox | None` purely to defer construction, re-narrowed with a bare `assert` at the use site** (`src/coder_eval/cli/evaluate_command.py:438`) — evaluate_command.py:404 declares `sandbox: Sandbox | None = None` and only builds it under `if not delegates_to_regrade:`, so the copy/adopt branch inside `_setup_and_run` needs `assert sandbox is not None  # built above whenever we reach this branch` (line 438), paired with `assert prior is not None` (line 421). The invariant is real today, but `assert` is the weakest available narrowing: it is stripped under `python -O`, and it encodes in a comment a relationship the type checker could hold structurally. Prefer hoisting the two mutually exclusive branches into two small functions (each taking a non-Optional `Sandbox` / `EvaluationResult`), or returning an explicit tagged result from the `delegates_to_regrade` decision, so the Optional never exists and the comment becomes unnecessary.
7. **[Axis 4] New `console.print(f"[red]✗ {e}[/red]")` renders untrusted recorded strings as Rich markup, letting a shared run directory hide or restyle parts of a grading error** (`src/coder_eval/cli/evaluate_command.py:477`) — The new handler added by this PR is `console.print(f"[red]✗ {e}[/red]")` (evaluate_command.py:477), and the shared `console` is a bare `rich.console.Console()` (cli/console.py) with markup enabled by default. The `RegradeError` messages routed here interpolate untrusted record fields — `_grade_in_container` builds `f"Task {task.task_id!r} ran under `driver: docker` ... records {recorded}"` with `recorded` containing the recorded `task_file` path (regrade.py:731-738) — so a `task_id` or `source_file` containing Rich tags is rendered as markup rather than text. Verified locally: `Console(force_terminal=True).print('[red]x ' + "...'[black on black]docker run evil/img[/]'..." + '[/red]')` emits `ESC[30;40m` and the payload is invisible. (The `--allow-recorded-commands` refusal itself is safe — it goes out through `typer.BadParameter`, which Click/Typer renders escaped.) Fix: `console.print(f"[red]✗ {escape(str(e))}[/red]")` using `rich.markup.escape`, and apply it to the three sibling handlers at lines 190, 195 and 409 that share the pattern. CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:N/I:L/A:N

## What's Missing

_Not produced — the synthesis agent for this section failed (individual spend limit reached during the run). This is **not** a finding of "nothing missing"; the pass did not run. Re-run the review to populate it._

_Not produced — the synthesis agent for this section failed (individual spend limit reached during the run). This is **not** a finding of "nothing missing"; the pass did not run. Re-run the review to populate it._

## Top 5 Priority Actions

1. Stop container grading from silently scoring 0.000 when `pre_run` seeded state outside the workspace: refuse by default when `task.pre_run` is non-empty (explicit opt-in, as `--allow-host-grading` already does) and, at minimum, write a durable `result.environment_info["graded_without_pre_run"]` stamp instead of the non-durable console warning at src/coder_eval/orchestration/regrade.py:764 — 3 of 10 in-tree docker tasks match this pattern and `execute` + `run --resume` reaches it with no flags.
2. Thread `recorded_task_file` through the in-container regrade branch (src/coder_eval/cli/run_task_internal_command.py:282, `_grade_recorded_run`, and `regrade_in_place`'s `Orchestrator(...)` call at src/coder_eval/orchestration/regrade.py:909), because today every container-graded row re-records `/work/task_dir/task.yaml` as `task_config.source_file`, which exists on no host and makes a later `evaluate <run_dir>` mount the wrong task dir or refuse outright.
3. Record the resolved image reference and digest into `environment_info` on both the run and the grade and warn (or refuse) when they differ, since `_build_image` (src/coder_eval/isolation/docker_runner.py:1319-1330) unconditionally rebuilds under the deterministic tag `coder-eval-task-<id>:built`, so a changed Dockerfile or base image grades the criteria against a different filesystem than the agent saw — and correct docs/USER_GUIDE.md:231, which promises a dispatch warning that no code emits.
4. Cover the new container-grading seams before merge: lines 377-401 of src/coder_eval/cli/run_task_internal_command.py are 0% covered and `orchestration/regrade.py` sits at 7.14%, and deleting `recorded_task=authored_task` (line 393) plus nulling `recorded_task_file` (line 235) both leave the full suite green — drive `_grade_recorded_run` end to end with `regrade_in_place` monkeypatched and assert the recorded task, the workspace mount guard, and that a container-graded row's `source_file` is the host path.
5. Close the three host-side safety and diagnosability gaps on the grading container: restore the pre-widening modes after `grant_container_access(self.grade_workspace, writable=True)` (src/coder_eval/isolation/docker_runner.py:694), which today leaves an operator-supplied `--workspace` world-writable forever; refuse or `O_NOFOLLOW` the `shutil.copy2` destination at src/coder_eval/orchestration/regrade.py:681 that three sibling writes in the same PR already guard with `is_symlink()`; and fold `docker.log` plus `grade.log` out of the scratch dir in a `finally:` so a failed grading container does not delete the very log its error message tells the operator to read (src/coder_eval/orchestration/regrade.py:792-827).

---

**Stats:** 1 🔴 · 5 🟠 · 10 🟡 · 7 🔵 across 8 axes reviewed.

uipreliga (approved):
Fix what you agree with and 🚢

PR #164, opened 2026-09-11
fix(timing): account for generation and tool time on every harness

uipreliga:
Per-task wall clock, LLM generation time and tool-execution time were only
correct for one harness. Measured across 4 `run.json` + 46 `task.json` files:

| harness | wall clock accounted for, before |
|---|---|
| claude-code | 97% |
| codex | 72% |
| antigravity | **15%** |

Antigravity recorded no generation time at all — the reducer read
`datetime.now()` once per flush and passed it as *both* message bounds with a
hardcoded `generation_duration_ms=0.0`, so every task page showed `0ms` of
generation and a `0%` thinking/tool/text breakdown, for months, with nothing
failing. Codex discarded the millisecond timestamps its SDK already delivers
and published the SDK item's own `duration_ms` instead: `0.0` for 70 of 211
sampled commands, absent for 25 more, and no execution bounds at all, so no
Codex tool call could be placed on a timeline.

## What changed

**The contract.** `AssistantMessage.generation_duration_ms` is now
`float | None`. `None` means *never measured*; `0.0` means *measured and took
no measurable time*. Every producer that used to conflate them is fixed —
including two the audit missed: Claude published a measured `0.0` for every
command force-closed without a tool result, and the simulator's trailing turn
reported `0s`, halving `avg_turn` in the HTML report for every simulation task.

**Generation time excludes tool execution.** Four of five harnesses interleave
tool calls into a single generation window, so the span between a message's own
bounds legitimately contains time the model did not spend generating. All four
now subtract the **union** of the closed tool intervals, clipped to the window,
through one shared helper. The union matters: these harnesses run tools
concurrently, and summing durations over-subtracts by exactly the overlap —
four concurrent 400ms calls inside a 1000ms window total 1600ms and clamp
generation back to the `0.0` this PR exists to remove.

**Attribution.** A mixed-kind emission's generation time and output tokens are
apportioned by one content-size weight vector. Previously all the time went to
the first block kind a priority chain tested (98.5% thinking on codex, 99.8% on
delegate), and the same output tokens were counted twice — once to thinking,
once to the tool — for 93% of Delegate's emissions.

**The page says what it cannot account for.** A new `Unaccounted` cell shows
wall clock minus generation minus tool execution, so the figures are displayed
against what they must reconcile to rather than each being individually
well-formed. The run-list `Duration` column now counts only rows that executed:
a codex nightly rendered "1300 tasks · 15h 29m" describing 397 tasks.

## Guardrails

- **CE058** — an unknown timing value may not become a numeric literal. Five
  syntactic forms, one id, including the `if x is None: x = 0.0` guard where a
  live instance was hiding and the `model_copy(update={...})` dict a
  keyword-only rule cannot see.
- **CE059** — an `AssistantMessage` may not take one clock read as both bounds.
  It exempts a call that passes `generation_duration_ms=None`: saying "no window
  was measurable" in the field built to say it is not a claim two stamps have to
  support.
- **`assert_timing_captured`** — a replay-based golden sensor. An AST rule cannot
  see that an SDK returned `0.0`; this runs the real reducer and asserts a
  resolved command carries both bounds and a duration, and that a turn which
  streamed a generation reports a positive window whose stamps actually span it.
- **Golden coverage for every built-in harness.** `antigravity`, `opencode` and
  `pi` had no recorded stream at all — the three whose timing was worst. The
  coverage test derives from `AgentKind` with an allowlist of exclusions, so the
  next harness cannot ship timing-blind.

## Numbers that move (correct, not regressions)

- **The cost simulator's thinking projection, for Codex and Delegate.** Both a
  smaller `thinkingMs` and a no-longer-double-counted `thinkingOutputTokens`
  feed it.
- **`avg_command_time_ms`.** A Claude command force-closed without a result now
  leaves *both* sides of the average instead of dragging it toward zero; and on
  Codex, `fileChange` and generic tool calls carry a duration they previously
  lacked, so the figure now covers every tool call rather than shell commands
  alone.
- **Codex `duration_ms` changed meaning** — the item's lifecycle (queueing and
  approval included) rather than the SDK's narrower command-run figure, which it
  deliberately overrides.

`docs/agents/HARNESS_PARITY.md` gains a `## Timing capture` section covering all
five harnesses, and names the two divergences left open (Antigravity's
orphan-poll wait, Delegate's missing bounds — both out of tree or deliberately
deferred).

## Verification

- `make verify` — 5209 passed, coverage 92.70%
- `make evalboard-verify` — 724 passed, `tsc --noEmit` + `next build` clean
- `make lint` — 555 (CE058/CE059 included)
- No task or experiment YAML changed; no new model, criterion, agent or config
  field, so no registry, merge-layer or `-D` surface is touched.

Reviewed phase by phase, then once across the whole diff. That last pass is what
caught OpenCode and Pi carrying the identical double-count — which no phase
covered and an earlier draft of the parity table wrongly denied.


bai-uipath (approved):
**Approve both.** The instrumentation is solid: the four-bucket identity closed to ~1ms on a live concurrent-tool task I ran against the stacked branch. The summary strip is the one thing worth a pass before merge.

## UI improvements

- **The strip flattens two levels into one row.** Five buckets that sum to the task wall clock sit beside a three-way split of one of them, which sums to Generation instead. Moving the per-kind split onto its own line would separate them.
- **"tool" means two unrelated things in the same row.** Time spent writing tool arguments, and time tools spent running. Renaming the sub-cell to "tool args" fixes it.
- **Unaccounted flips sign when tools overlap.** It subtracts the sum of per-tool durations where the identity uses the union, so concurrent tools get counted twice. On a task with two concurrent sleeps it rendered -615ms amber where the honest answer is +2.5s of sandbox setup and grading.
- **The execution bounds that needs are already in task.json,** and already in scope where the per-tool duration is set, so it should be a small change.

## Minor, fix if you agree

- **No parity test covers the TS decomposition against the Python one.** Every current TS test asserts against the sum, so the suite agrees with the bug above.
- **opencode and pi don't subtract still-open tools** the way antigravity and codex do, though the argument applies to them equally now their windows tile.
- **The collector's head/tail measurement doesn't exclude sub-agent messages,** unlike its two sibling call sites.
- **Delegate in coder_eval_uipath keeps both defects this removes,** while sitting outside the new rules' scope.
- **#165 hasn't run the gating workflows,** since its base is a feature branch. Verify, lint, pyright and the evalboard suite were all green locally against its head.

uipreliga:
Thanks — every point landed. The five for this PR are in `7483b67f` (which also rebases the branch onto `main`); the sixth is on the `#165` branch.

## The strip

**Unaccounted flips sign when tools overlap.** Confirmed and fixed. You were right that the bounds were already in reach: `CommandTelemetry.execution_started_at` / `execution_completed_at` are in `task.json`, and the evalboard already had the `CommandEntry` in scope where it sets `durationMs`. They now travel onto `MessageToolUse` as `execStartMs` / `execEndMs`, and the cell unions the intervals instead of summing durations. A call the harness timed but did not *bound* still contributes its own `durationMs`, so no run silently loses tool time to the change.

**Two levels in one row.** The per-kind split moved to its own row under a `Generation split` label. The top row's Generation + Tool exec + Unaccounted sum to the wall clock; the bottom row sums to Generation. The caveat about apportionment moved with the split, since it qualifies that and not the measured total above it.

**"tool" naming.** Now `tool args`, one row below `Tool exec`. The bare word did name both.

## Minor

**No parity test.** Added, and pointed at the root of the divergence rather than at the rendered cell: `busyMs` (TS) and `busy_ms` (Python) answer the same question about the same `task.json`, so neither owns the numbers — `tests/_fixtures/timing_union_cases.json` does, and both suites replay all 14 cases. The Python half also fails if the TS half stops reading the file, so a corpus only one side replays can't survive. On top of that, `toolExecutionMs` has its own cases for the concurrent, sequential, cross-message, unbounded-fallback and never-timed shapes.

One implementation note: the helpers live in `evalboard/lib/timing.ts`, not `runs.ts`. `runs.ts` reads `node:fs`, so a client component importing a *value* from it fails the Next.js build — only types may cross that boundary.

**OpenCode and Pi don't subtract still-open tools.** Fixed both, mirroring Antigravity and Codex. Agreed the argument applies equally now their windows tile: without it the part of a straddling call that ran before the boundary is published as generation while the call's own `duration_ms` counts it again.

**The collector's head/tail doesn't exclude sub-agent messages.** Fixed on the `#165` branch — `_overhead_ms` now filters `parent_tool_use_id is None`, which is the rule `codex_agent._token_usage_from_messages` and `scripts/timing/decompose_run.py` already apply. It matters in both directions: the identity sums generation over the main thread only, so bracketing the span with a sub-agent message shrinks the head or the tail by time no bucket then claims, and Codex's recovered child messages carry the *child's* clock. Mutation-verified — dropping the filter fails both new cases. It'll appear on that branch when I push its rebase onto the new base.

## Not fixed

**Delegate in `coder_eval_uipath`.** Acknowledged, and left alone: it's a separate repo, so it needs its own PR. Worth noting the exposure is real — it sits outside CE058/CE059's scope, so nothing will catch a reintroduction there either.

**#165 hasn't run the gating workflows.** Still true while its base is a feature branch. GitHub retargets it to `main` when this merges, at which point they run for real.

## Also in this push

The three CodeQL alerts (a module imported with both `import` and `from … import` in the agent test modules) are cleared — the local imports are hoisted.

## Verification

- `make verify` — 5239 passed, 92.71% coverage
- `make lint` — 555
- evalboard — `tsc` clean, 749 tests, `next build`

PR #166, opened 2026-09-11
feat(harbor): task.yaml/experiment.yaml → Harbor export + coder-eval as a Harbor agent

akshaylive:
## Summary

Adds a Harbor-interop layer for coder-eval:

- **Export** (`coder-eval export --format harbor`): converts a coder-eval `task.yaml` (optionally × `experiment.yaml` variants) into a Harbor task directory, using coder-eval's own criteria as the grader.
- **coder-eval as a Harbor agent** (`CoderEvalAgent`, `-a coder_eval.harbor.agent:CoderEvalAgent`): runs `coder-eval execute --format harbor` inside a Harbor trial container, in-place at the container's own `WORKDIR` (`--workspace-dir`), and hands Harbor back a real ATIF `trajectory.json`.
- Supporting fixes found only by running the whole pipeline against a real `harbor==0.22.0` install, real Docker, and real Bedrock credentials: `[environment.env]`/`[verifier.env]` auto-derived from `sandbox.docker.env_passthrough` (`${VAR:-}` templates so Harbor doesn't hard-require every credential var), an always-written `environment/Dockerfile`, a fixed placeholder `instruction.md`, `template_sources` directories copied into the export with rewritten paths, and `TaskDefinition` no longer requiring a placeholder success criterion.
- A new **Harbor E2E workflow** (`.github/workflows/harbor-e2e.yml`) that exports and runs 4 real scenarios (plain docker driver, llm_judge, custom BYOD image, template_sources) end-to-end and asserts `reward.json == 1.0` + `trajectory.json` + `task.json` presence.

## Test plan

- [x] `make format && make check && make lint` clean
- [x] `make typecheck` at the known 9-error baseline (pre-existing optional-extra imports, unchanged)
- [x] `make test` — 5069 passed, same 10 pre-existing unrelated `litellm`-extra failures
- [x] All 4 Harbor E2E scenarios verified locally end-to-end against real Docker + real Bedrock credentials (reward 1.0, no exceptions, for each)
- [ ] `harbor-e2e.yml` triggered via `workflow_dispatch` on this PR's branch to confirm it passes in CI too (not triggered by `pull_request` itself — see the workflow's own comment)


uipreliga (commented):
# Review: coder_eval — pr:166

Scope: pr:166 · branch `akshaya/harbor_experiment_export` · `0420491` · 2026-09-11T20:50Z · workflow variant

**Change class:** complex — introduces a new public interop layer (Harbor export + coder-eval as a Harbor agent) with new CLI commands, subprocess/container invocation, an ATIF trajectory schema mapping, and a relaxed TaskDefinition validator; correctness requires reasoning about control flow, schema semantics and the grading contract

The new Harbor interop layer is architecturally sound and cleanly factored (Architecture 8.5, API Surface 9.2, with a golden-fixture round trip and a real portability audit), but the genuine risks are score-integrity ones — a deleted `success_criteria` invariant that lets `evaluate` and `run --resume` grade a criteria-less task as SUCCESS at 0.000 and exit 0, a reward writer that publishes a grading-time infrastructure crash to Harbor as a measured `reward: 0.0`, and an emit→hydrate/export path that silently drops `result_status`, `run_limits`/`checker_context`/`simulation` and dataset `${row.*}` expansion so identical agent output scores differently on the Harbor path — compounded by a hand-vendored ATIF mirror that rejects both of harbor 0.22.0's documented sub-agent ref shapes, an unquoted task-YAML `working_dir` interpolated into the generated verifier shell script, reference/template copies that dereference host symlinks into a distributable bundle, two brand-new CLI commands with zero CliRunner coverage, and an `export -e` path that exports nothing and exits 1 for the archetypal model A/B; bottom line: the design is right, the correctness plumbing is not, and this should not merge until the verdict-affecting drops and the grading-guard bypass are closed.

## Summary

| 1. Code Quality & Style | 7.1 / 10 | 0 | 1 | 3 | 4 | `_unhonorable_override_reason` skips every agent-override variant citing an adapter this PR builds — the archetypal model A/B `export -e experiment.yaml` emits zero directories and exits 1 |
| 2. Type Safety | 5.4 / 10 | 1 | 1 | 1 | 1 | Removing `TaskDefinition.validate_success_criteria` was replaced by a guard on only the `run --to_run` path, so `evaluate` and `run --resume`'s `to_grade` set grade a criteria-less task as SUCCESS at weighted_score 0.0 (exit 0), with no test and two stale doc lines |
| 3. Test Health | 5.7 / 10 | 0 | 2 | 4 | 3 | Both new top-level CLI commands ship with their entire body untested (`export` 17.54%, `harbor reward` 57.14%) — the PR's headline entry point has zero CliRunner coverage |
| 4. Security | 8.4 / 10 | 0 | 1 | 1 | 1 | Unquoted interpolation of task-YAML `sandbox.docker.working_dir` into the generated Harbor verifier shell script (`tests/test.sh`) |
| 5. Architecture & Design | 8.5 / 10 | 0 | 0 | 3 | 0 | New run-directory artifact `trajectory.json` ships as three bare filename literals with no `path_utils` constant (CE053 structurally blind) and is undocumented in the run-layout contract |
| 6. Error Handling & Resilience | 6.8 / 10 | 0 | 2 | 2 | 2 | reward.py's unmeasured guard tests only `weighted_score is None`, so a grading-time INFRASTRUCTURE failure is published to Harbor as a measured `{"reward": 0.0}` |
| 7. API Surface & Maintainability | 9.2 / 10 | 0 | 0 | 1 | 3 | `--trajectory` without `--format harbor` is silently discarded: trajectory-dependent criteria score 0.0 and the run reports an ordinary criterion failure |
| 8. Evaluation Harness Quality | 6 / 10 | 0 | 4 | 0 | 0 | Host-reachable `--workspace-dir` runs in-place in a directory that is never cleared while also suppressing the stale-artifact warning (orchestrator.py:1657), so a prior run's files satisfy file-based criteria; the new guard at batch.py:120-131 checks only task count and driver |

**Overall Score**: 7.1 / 10 · **Weakest Axis**: Type Safety at 5.4 / 10
**Totals**: 🔴 1 · 🟠 11 · 🟡 15 · 🔵 14 across 8 axes.

> **Note on length:** the fully-rendered review (every finding with its complete evidence chain and adversarial-verification note) is 142 KB — over GitHub's 64 KB comment limit. Findings below are abridged to title + location + the fix; the unabridged text for each lives in the per-axis report files named at the end of each section.

## Blockers

1. 🔴 **Critical [Axis 2]** Removing `TaskDefinition.validate_success_criteria` was replaced by a guard on only the `run --to_run` path, so `evaluate` and `run --resume`'s `to_grade` set grade a criteria-less task as SUCCESS at weighted_score 0.0 (exit 0), with no test and two stale doc lines
   - `src/coder_eval/models/tasks.py:495` · *cross-axis: 1, 2, 3, 5, 6, 7, 8*
   - This PR deleted the model-level guard (`git diff origin/main...HEAD -- src/coder_eval/models/tasks.py` shows the whole validator removed after line 792): [code/transcript — see per-axis file] so `success_criteria: list[SuccessCriterion] = MergeField(...)` (tasks.py:495) now accepts `[]`. Verified: `TaskDefinition(..., success_criteria=[])` validates (omitting the key entirely is still rejected as `Field required`, so the hole is the explicit empty list — reachable from a typo, a bad merge, or a `-D` override that clears the list, exactly as the replacement guard's own docstring says). The compensating guard `_reject_empty_criteria_under_grade` (`cli/run_command.py:983`) has exactly ONE call site, `cli/run_command.py:1140`, and its own docstring claims a scope it does not have: "Refuse a task with zero ``success_criteria`` under ``run``/``evaluate``". Two graded paths bypass it. (a) `coder-eval evaluate` never calls it — `grep -rn "_reject_empty_criteria_under_grade" src/` returns only run_command.py:983/1140 plus the tests. Reproduced live against this HEAD with a 6-line task YAML carrying `success_criteria: []`: [code/transcript — see per-axis file] That is precisely the outcome the guard exists to prevent…
   - Full analysis: `02-type-safety.md`
   - Theme group (3 members): `src/coder_eval/models/tasks.py:495`, `src/coder_eval/cli/run_command.py:983`, `src/coder_eval/cli/run_command.py:1140`
2. 🟠 **High [Axis 1]** `_unhonorable_override_reason` skips every agent-override variant citing an adapter this PR builds — the archetypal model A/B `export -e experiment.yaml` emits zero directories and exits 1
   - `src/coder_eval/harbor/experiment_packager.py:84` · *cross-axis: 1, 5, 8*
   - The agent branch of `_unhonorable_override_reason` is live code whose entire justification is false as of this PR, and it blocks the feature the PR advertises. Its docstring (line 63) says Harbor's agent phase is "Harbor's own (or the unbuilt C1.2 coder-eval-as-agent embed adapter...)" and its user-facing message (line 88) says "Building the C1.2 coder-eval-as-agent embed adapter (tmp/harborframework.md, Part A step 8) is the fix; until then this variant is skipped" — but `src/coder_eval/harbor/agent.py:1` is titled "``CoderEvalAgent`` — coder-eval as a Harbor agent (C1.2)" and `packager.py:506-508` already carries `task.agent` verbatim into `environment/task.yaml` (`"agent": (task.agent.model_dump(mode="json", exclude_none=True) if task.agent is not None else {"type": "claude-code"})`), which `CoderEvalAgent.run()` then executes. So the agent override IS honored now, yet it is still refused. Reproduced live (not inferred): exporting a plain docker task with `-e experiments/default.yaml` yields `exported: 0 skipped: 1` with reason `variant 'default' sets an agent override (source: experiment-defaults) ...`, because `experiment_packager.py:160-161` sets `default_experiment = experiment` when the experiment IS…
   - Full analysis: `01-code-quality.md`
   - Theme group (2 members): `src/coder_eval/harbor/experiment_packager.py:88`, `src/coder_eval/harbor/experiment_packager.py:84`
3. 🟠 **High [Axis 2]** The hand-vendored ATIF mirror diverges from the `harbor==0.22.0` this PR pins: `SubagentTrajectoryRef` (atif_models.py:74) REJECTS both of harbor's documented ref shapes other than its own (`session_id` present, and the spec's `trajectory_path`-only "file-ref form") under `extra="forbid"` + required `trajectory_id`,…
   - `src/coder_eval/harbor/atif_models.py:74`
   - `atif_models.py`'s docstring says it "Mirrors the schema in ``harbor.models.trajectories`` (harbor 0.20.0)" and `tests/test_atif_models.py:5` says "CI never installs harbor". Both are now false: this PR adds `harbor = ["harbor==0.22.0"]` to `pyproject.toml` and makes both pyright jobs install `--extra harbor`. I installed `harbor==0.22.0` and diffed `model_fields` field-by-field against the vendored classes. Four real divergences, two of them load-bearing: 1. **Rejects a valid document.** `class SubagentTrajectoryRef` (atif_models.py:74) declares only two fields — `trajectory_id: str = Field(...)` (line 79, REQUIRED) and `trajectory_path: str | None` (line 80) — under `model_config = ConfigDict(extra="forbid")` (line 77). Real harbor 0.22.0 has FOUR fields and makes the first optional: `trajectory_id: str | None = None`, `session_id: str | None = None`, `trajectory_path: str | None = None`, `extra: dict[str, Any] | None = None`. Reproduced on this HEAD with a trajectory whose subagent ref carries `session_id`: [code/transcript — see per-axis file] That kills both read paths on a legitimate input: `run_evaluation`'s `Trajectory.model_validate_json(trajectory.read_text(...))` (`cli/evaluate_command.py:424`) exits 1 with "Could not…
   - Full analysis: `02-type-safety.md`
4. 🟠 **High [Axis 3]** Both new top-level CLI commands ship with their entire body untested (`export` 17.54%, `harbor reward` 57.14%) — the PR's headline entry point has zero CliRunner coverage
   - `src/coder_eval/cli/export_command.py:79`
   - Two members of one mechanical class, scored once. (a) `src/coder_eval/cli/export_command.py` is at 17.54% — the entire body, lines 79-123, is uncovered, i.e. everything after the signature: the `if format not in _SUPPORTED_FORMATS:` refusal (79-81), the `"[red]✗[/] Without --experiment, pass exactly one task YAML."` guard (84-86), the `except (TaskNotExportableError, CriteriaNotExportableError)` → `raise typer.Exit(1)` mapping (89-91), the whole experiment branch (97-121), and the failure gate `if not exp_result.exported:\n raise typer.Exit(1)` (122-123). `grep -rn "export_command" tests/` returns nothing and there is no CLI-registration/`--help` test anywhere in `tests/`, so `coder-eval export` — the PR's headline user-facing command — is never invoked through `CliRunner` at all; only the library functions `export_task`/`export_experiment` are tested. This is not theoretical: the path-traversal guard this same PR added (`UnsafeExportPathError`, raised at `harbor/experiment_packager.py:129` from `_out_subdir`, which sits OUTSIDE the `try` at line 200) is caught by nobody, and `uv run coder-eval export t.yaml -e e.yaml -o out` with `variant_id: "../../../../tmp/pwned"` prints a full Rich traceback ending `UnsafeExportPathError:…
   - Full analysis: `03-test-health.md`
5. 🟠 **High [Axis 3]** `harbor/atif_hydrate.py` at 60.71% — the documented sub-agent flattening block and the multimodal-message branch have zero coverage, while a fixture that exercises both already exists and is never fed through it
   - `src/coder_eval/harbor/atif_hydrate.py:118`
   - `atif_hydrate.py` is the lowest-covered new source module in the PR (60.71%, 18 of 58 statements missed). Two uncovered regions are documented contract, not defensive filler. (a) Lines 119-134 — the entire sub-agent flattening loop opened by `for sub in trajectory.subagent_trajectories or []:` at line 118, whose own docstring (lines 102-107) promises "each embedded child's tool calls are appended onto whichever main-thread turn contains the spawning tool call (matched by ``tool_call_id``), falling back to the last turn when no spawning call is found (an orphaned embed)". Neither the match branch (`target = turn; break`, 131-132) nor the orphan fallback (`target = turns[-1] if turns else None`, 128) is ever executed. (b) Line 62 — `return "\n".join(part.text for part in step.message if part.type == "text" and part.text)`, the `list[ContentPart]` arm of `_step_text`. `Step.message` is typed `str | list[ContentPart]` (`harbor/atif_models.py:184`), and hydrate's whole purpose is consuming trajectories produced OUTSIDE this process, where the list form is legal ATIF — but every test feeds a trajectory built by our own `atif_emit`, which only ever emits `str`, so the external-producer shape is never exercised. Because these commands…
   - Full analysis: `03-test-health.md`
6. 🟠 **High [Axis 4]** Unquoted interpolation of task-YAML `sandbox.docker.working_dir` into the generated Harbor verifier shell script (`tests/test.sh`)
   - `src/coder_eval/harbor/packager.py:87`
   - `_TEST_SH_TEMPLATE` interpolates a task-YAML-controlled value straight into a `/bin/sh` script with nothing but surrounding double quotes: packager.py:87 `coder-eval evaluate /tests/task.yaml "{workdir}" --in-place --run-dir /logs/verifier || true` packager.py:520 `path.write_text(_TEST_SH_TEMPLATE.format(workdir=workdir), encoding="utf-8")` `workdir` comes from `docker_cfg.working_dir` (packager.py:282 `workdir = docker_cfg.working_dir or _find_workdir(dest_dockerfile) or DEFAULT_WORKDIR`, and packager.py:312 `workdir = docker_cfg.working_dir`). The only validator on that field (`models/sandbox.py:295 _validate_working_dir`) checks `if not v.startswith("/")` and the reserved-path set — it does not reject quotes, `$(`, backticks or newlines. Inside `"..."` in sh, `$(...)` and `` ` `` still expand, and a `"` closes the quote outright. VERIFIED by running `export_task` on a task with `working_dir: '/app"; id > /tmp/pwned_verifier; echo "'`; the emitted `tests/test.sh` reads verbatim: `coder-eval evaluate /tests/task.yaml "/app"; id > /tmp/pwned_verifier; echo "" --in-place --run-dir /logs/verifier || true` That is arbitrary command execution in Harbor's verifier phase, in a container whose environment this same packager populates…
   - Full analysis: `04-security.md`
   - CVSS: `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H`
7. 🟠 **High [Axis 6]** reward.py's unmeasured guard tests only `weighted_score is None`, so a grading-time INFRASTRUCTURE failure is published to Harbor as a measured `{"reward": 0.0}`
   - `src/coder_eval/harbor/reward.py:65`
   - The module docstring states the contract this line fails to implement: "An unmeasured row ... must not become `reward=0.0`: that would train 'the agent's behaviour was bad' from a measurement that never happened." But the guard is only: [code/transcript — see per-axis file] `weighted_score is None` is written ONLY by `Orchestrator._finalize_weighted_score`'s `grade=False` arm (orchestrator.py:1084). A grading pass that CRASHES takes the `self.grade` arm instead and calls `EvaluationResult.calculate_weighted_score`, whose first statement short-circuits an empty results list to a hard zero (models/results.py:661-663): `if not self.success_criteria_results or not criteria: self.weighted_score = 0.0; return`. So an ERROR row carries `weighted_score = 0.0`, not `None`. That shape is reachable by design, not by accident: `criteria/base.py:87` declares `_ESCALATING_EXCEPTIONS = (JudgeInfrastructureError, CheckerMisuseError)` — exceptions that deliberately propagate OUT of the checker instead of being captured into a scored-0.0 result, precisely because "a judge-infra outage or a checker-contract misuse is not an agent failure". They escape `check_all_async`, hit the Orchestrator's broad handler, and finalize the row as…
   - Full analysis: `06-error-handling.md`
8. 🟠 **High [Axis 6]** `evaluate --format harbor` keys write-back/restore on `prior is not None` with no `target.mode` test — writes a spurious `task.json` into the graded work dir, prints a false "original kept as task.execute.json" notice, and wedges later `evaluate` runs into RUN_DIR mode
   - `src/coder_eval/cli/evaluate_command.py:615` · *cross-axis: 6, 7*
   - The new `--format harbor` branch seeds a synthetic prior (`prior = seed_from_atif_trajectory(...)`, line 428) on the WORK_DIR shape. Three downstream sites key on `prior is not None` alone, with no `target.mode` test: [code/transcript — see per-axis file] `_write_back` (line 640) is documented as "Replace the graded **run's** `task.json`" and writes `run_dir / TASK_JSON_FILENAME`. In WORK_DIR mode `target.target` is the directory being graded. Reproduced with `coder-eval evaluate <task.yaml> <workdir> --format harbor --trajectory t.json --run-dir run`: exit 0, and `WORKDIR CONTENTS: ['task.json', 'test.txt']` — plus the console claim "Updated .../workdir/task.json (original kept as task.execute.json); run `coder-eval aggregate` to refresh run.json", where no `task.execute.json` exists and the directory is not a run. Consequences, all mechanical: - Under Harbor the work dir is the container WORKDIR that Harbor syncs back, so a coder-eval run record appears among the agent's deliverables. - The command is not re-runnable: `evaluate_target.py:105` reads a second positional holding `task.json` as `EvaluateMode.RUN_DIR`, and the record just written parses, so the mode sticks — after which the new guard at evaluate_command.py:411…
   - Full analysis: `06-error-handling.md`
   - Theme group (1 members): `src/coder_eval/cli/evaluate_command.py:620`
9. 🟠 **High [Axis 8]** Host-reachable `--workspace-dir` runs in-place in a directory that is never cleared while also suppressing the stale-artifact warning (orchestrator.py:1657), so a prior run's files satisfy file-based criteria; the new guard at batch.py:120-131 checks only task count and driver
   - `src/coder_eval/orchestration/batch.py:120`
   - `Orchestrator.workspace_dir` existed before this PR but had exactly one writer — `cli/run_task_internal_command.py:310`, the in-container entry point, where the WORKDIR is a fresh container filesystem every run, so stale leakage was structurally impossible (`git grep -n "workspace_dir=" origin/main -- src/coder_eval/` returns only that line plus a log message). This PR promotes it to a host-reachable CLI flag on both `run` and `execute`, where the named directory persists across invocations. The new guard at `orchestration/batch.py:120-130` validates two things — `if len(resolved_tasks) != 1: raise ValueError(...)` and `if _workspace_dir_task.task.sandbox.driver == "docker": raise ValueError(...)` — and nothing about staleness. Meanwhile `clear_rerun_artifacts` (`orchestration/batch.py:452-458`) only removes `rt.run_dir / "artifacts" / rt.task.task_id`, never the workspace dir, and `orchestrator.py:1656-1659` explicitly turns the stale-artifact warning OFF in exactly this mode: `if (self.workspace_dir is None and direct_target is not None and direct_target.exists() and any(direct_target.iterdir())):` — a suppression that was correct for the fresh-container-only writer and is wrong for a host directory. Verified live. A workspace…
   - Full analysis: `08-harness-quality.md`
10. 🟠 **High [Axis 8]** Harbor export writes task YAMLs from hand-picked untyped `dict[str, object]` payloads that never round-trip through `TaskDefinition`, silently dropping score-affecting fields (`run_limits`, `checker_context`, `simulation`) with no completeness check
   - `src/coder_eval/harbor/packager.py:503` · *cross-axis: 2, 8*
   - Both emitted task YAMLs are built from a hardcoded key list with no completeness check against `TaskDefinition.model_fields`: `_write_agent_phase_task_yaml`'s payload at `packager.py:503-511` is `{task_id, description, agent, sandbox, success_criteria}` (+ `initial_prompt`), and `_write_verifier_task_yaml`'s at `packager.py:398-404` is `{task_id, description, agent, initial_prompt, success_criteria}` (+ `reference`). Everything else is dropped in silence. The export DOES warn for `pre_run` and `post_run` (`packager.py:174-185`) — the two fields whose loss cannot change a score — so the warning set is precisely inverted from the risk. What is dropped, verified live by exporting a task carrying `run_limits: {max_turns: 3, max_usd: 0.25, turn_timeout: 60}` and an enabled `simulation:` block: $ coder-eval export tmp/ax8/sim.yaml -o tmp/ax8/simexport -> `✓ Exported ... (format: harbor)`, exit 0 environment/task.yaml contains NO `run_limits:` and NO `simulation:` key at all; task.toml contains no `[agent] timeout_sec` and no `[verifier] timeout_sec` either (those two are written only from `run_limits.task_timeout`, `packager.py:539-555`). 1. `run_limits` — only `task_timeout` has a Harbor home. `max_turns`, `turn_timeout`, `max_usd`,…
   - Full analysis: `08-harness-quality.md`
   - Theme group (1 members): `src/coder_eval/harbor/packager.py:503`
11. 🟠 **High [Axis 8]** ATIF emit->hydrate round-trip drops `CommandTelemetry.result_status` (and `duration_ms`), so `command_executed(require_success: true)` scores 0.0 on the `evaluate --format harbor` path for a command that succeeded (verified: 1.0 native vs 0.0 hydrated)
   - `src/coder_eval/harbor/atif_hydrate.py:48`
   - `atif_emit` preserves the field — `atif_emit.py:112-113`: `if cmd.result_status is not None: call_extra["result_status"] = cmd.result_status` — but `atif_hydrate._tool_call_to_command` never reads `call.extra` back. Its `CommandTelemetry(...)` construction at `atif_hydrate.py:48-56` sets only `tool_name`, `tool_id`, `timestamp`, `parameters`, `result_summary`, `assistant_turn_index`, `sequence_number`; `grep -n "extra" src/coder_eval/harbor/atif_hydrate.py` returns nothing. `CommandTelemetry.result_status` (`models/telemetry.py:372`) therefore defaults to None, and `criteria/command_executed.py:173` skips every such command: `if criterion.require_success and cmd.result_status != "success": continue`. Verified round-trip: emitted ToolCall.extra = {'result_status': 'success'} hydrated result_status = None (one `CommandTelemetry(tool_name="Bash", result_status="success", ...)` through `evaluation_result_to_trajectory` then `trajectory_to_turn_records`). So the same agent trajectory scores 1.0 under a native `coder-eval run` and 0.0 under `coder-eval evaluate --format harbor --trajectory trajectory.json` — the Axis-8 blocker class, on the very criterion family the hydrate module exists to serve (`atif_hydrate.py:6-9` names…
   - Full analysis: `08-harness-quality.md`
12. 🟠 **High [Axis 8]** Exporting a dataset-backed task silently emits one Harbor task whose prompt and criteria still contain literal `${row.<field>}` placeholders (no `dataset` guard in `export_resolved_task`)
   - `src/coder_eval/harbor/packager.py:166`
   - `export_task` calls `load_task` (`packager.py:137`), which does NOT run `expand_dataset` — dataset fan-out happens later, in the experiment pipeline. `export_resolved_task`'s refusal block checks the criteria audit (`packager.py:162`) and the sandbox driver (`packager.py:166`: `if task.sandbox.driver != "docker":`) and nothing else; `grep -rn "dataset" src/coder_eval/harbor/` finds the word only in docstrings, never in a guard. Verified live on a two-row dataset task whose prompt is `Write ${row.expected} to out.txt` and whose criterion is `file_contains: {path: out.txt, includes: ["${row.expected}"]}`: $ coder-eval export tmp/ax8/ds.yaml -o tmp/ax8/dsexport -> exit 0, only an unrelated WORKDIR warning environment/task.yaml: `initial_prompt: Write ${row.expected} to out.txt` tests/task.yaml: `includes:` / `- ${row.expected}` and `description: contains ${row.expected}` Both rows are gone and the placeholders are unsubstituted, so the exported task asks the agent for a literal `${row.expected}` and grades against that string. The reward is then either a false 1.0 (agent takes the prompt literally) or a measured 0.0 for a task that was never expressible — in both cases Harbor records it as a real measurement. A second, related…
   - Full analysis: `08-harness-quality.md`

1. 🟡 **Medium [Axis 1]** `--format` is a free-string magic-value layer: 6 literals across 3-4 commands, three hand-rolled one-value checks with two divergent exit codes, no shared constant/enum or `click.Choice`
   - `src/coder_eval/cli/run_command.py:472` · *cross-axis: 1, 2, 7*
   - The `"harbor"` literal is hand-compared at five independent sites with no shared constant and three different refusal mechanisms, so a second format means editing all of them: `cli/run_command.py:472` `if format is not None and format != "harbor": raise typer.BadParameter(f"Unsupported --format {format!r}. Supported: harbor.")` (exit…
   - Full analysis: `01-code-quality.md`
   - Theme group (2 members): `src/coder_eval/cli/run_command.py:358`, `src/coder_eval/cli/run_command.py:472`
2. 🟡 **Medium [Axis 1]** `write_trajectory_json_strict` is an abstraction for a single call site, justified by a caller ("the report backfill") that does not exist anywhere in the repo
   - `src/coder_eval/harbor/atif_emit.py:366`
   - `write_trajectory_json_strict` (line 366) is public, has no `__all__` gating it, and is called from exactly one place — its own sibling wrapper `write_trajectory_json` at line 392 (`return write_trajectory_json_strict(result, path)`). Its docstring justifies the split with "for callers that must distinguish 'nothing to emit' from…
   - Full analysis: `01-code-quality.md`
3. 🟡 **Medium [Axis 1]** Cyclomatic-complexity cluster: 4 new and 3 worsened functions in the CC 10–20 band across the new harbor package and the CLI
   - `src/coder_eval/harbor/atif_emit.py:172`
   - Theme-grouped, scored once. Verified with `uv run radon cc -s -n C` on the PR HEAD and on `git show origin/main:src/coder_eval/cli/run_command.py` for the deltas. NEW: `src/coder_eval/harbor/atif_emit.py:172 _emit_turn` C(19) — the worst; it interleaves command-by-index grouping, a `nonlocal first_step_of_turn` closure `_step_extra`, an…
   - Full analysis: `01-code-quality.md`
4. 🟡 **Medium [Axis 2]** `harbor/agent.py:52-56` re-introduces 3 local pyright errors (6 total, up from a pre-existing 3): commit 6c47e1f6's scoped `# pyright: ignore` comments were deleted in c2b79d7d in favor of a CI-only `--extra harbor`, while `make install` (Makefile:15) still syncs only `dev`+`uipath` and claims "parity with CI"
   - `src/coder_eval/harbor/agent.py:52`
   - `uv run pyright` on this HEAD reports 6 errors, 3 of them new and all in this file: [code/transcript — see per-axis file] History matters here: commit 6c47e1f6 ("fix(harbor): suppress pyright reportMissingImports for the harbor package") added a scoped `# pyright: ignore[reportMissingImports]` to each of those three lines, matching the…
   - Full analysis: `02-type-safety.md`
5. 🟡 **Medium [Axis 3]** The flat `run_dir/trajectory.json` promotion — the hinge binding `execute --format harbor` to `CoderEvalAgent` — executes in a test but is asserted by nothing, and the multi-task branch is untested
   - `src/coder_eval/cli/run_command.py:693`
   - `run_command.py:692-695` is the contract between the PR's two halves: `flat_trajectory_path = run_dir / "trajectory.json"` / `if len(written) == 1 and written[0] != flat_trajectory_path:` / `await asyncio.to_thread(shutil.copy2, written[0], flat_trajectory_path)`. On the read side, `harbor/agent.py:139` does `trajectory_path =…
   - Full analysis: `03-test-health.md`
6. 🟡 **Medium [Axis 3]** `harbor/atif_emit.py` reassembles the ATIF record field-by-field with no full-field parity test — five mapped fields have no assertion anywhere, so a future `TurnRecord`/`AssistantMessage`/`CommandTelemetry` field drops silently
   - `src/coder_eval/harbor/atif_emit.py:96`
   - `atif_emit` is the event-reassembled-record shape the axis brief singles out: a persisted artifact rebuilt field-by-field from coder-eval's own models, where an unmirrored field vanishes with no failing assertion. `grep -rn model_fields tests/test_atif_emit.py tests/test_atif_hydrate.py tests/test_atif_models.py` returns nothing — there…
   - Full analysis: `03-test-health.md`
7. 🟡 **Medium [Axis 3]** `BatchRunConfig.workspace_dir`: the >1-task refusal arm (`src/coder_eval/orchestration/batch.py:121-125`) has zero tests, and no PR-gated test proves the CLI → `BatchRunConfig` → `run_batch` → `Orchestrator(workspace_dir=…)` forwarding (`batch.py:203`); only the docker-driver arm is tested
   - `src/coder_eval/orchestration/batch.py:121`
   - The PR adds `workspace_dir` to `BatchRunConfig` (`orchestration/config.py:120`) and two resolution-time guards at the top of `run_batch` (`orchestration/batch.py:120-131`), then plumbs it to `Orchestrator(..., workspace_dir=config.workspace_dir)` at `batch.py:203`. Only the second guard is tested. `grep -rn workspace_dir tests/` shows…
   - Full analysis: `03-test-health.md`
8. 🟡 **Medium [Axis 3]** Packager/experiment-packager mapping edge cases untested: branch coverage confirms packager.py 209-213 (both non-`initial_prompt` prompt arms), 356 (`docker image inspect` non-zero exit), 512->514 (agentless payload shape), and experiment_packager.py 94 (simulation skip reason) / 122 (dataset `row_id` segment) are…
   - `src/coder_eval/harbor/packager.py:209`
   - Six same-class gaps in the export mapping, enumerated so they can be closed in one pass. (1) `packager.py:209-213` — `_resolve_prompt_text` has three arms and only `task.initial_prompt` is tested; `return (task_file.parent / task.initial_prompt_file).read_text(encoding="utf-8")` (210-211) and the agentless fallback `return…
   - Full analysis: `03-test-health.md`
9. 🟡 **Medium [Axis 4]** Harbor export copies reference/template trees with symlinks dereferenced and no containment check, pulling host secrets into a distributable bundle
   - `src/coder_eval/harbor/packager.py:531` · *cross-axis: 4, 6*
   - Three copy sites in the packager use a bare `shutil.copytree` (default `symlinks=False`, i.e. symlinks are DEREFERENCED and their target's content written as a regular file), and the reference source is joined with no containment check: packager.py:527 `source = task_file.parent / task.reference.directory` packager.py:531…
   - Full analysis: `04-security.md`
   - CVSS: `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:N/A:N`
10. 🟡 **Medium [Axis 5]** New run-directory artifact `trajectory.json` ships as three bare filename literals with no `path_utils` constant (CE053 structurally blind) and is undocumented in the run-layout contract
   - `src/coder_eval/harbor/atif_emit.py:418`
   - The PR adds a new run-record file to the run directory in two locations and names it as a bare string at every one of the three cross-package sites: producer `harbor/atif_emit.py:418` `trajectory_path = task_json.with_name("trajectory.json")`; promoter `cli/run_command.py:692` `flat_trajectory_path = run_dir / "trajectory.json"`; reader…
   - Full analysis: `05-architecture.md`
11. 🟡 **Medium [Axis 5]** The 2173-line `harbor/` package and its two new CLI commands (`export`, `harbor reward`) are absent from every architecture/docs surface (zero doc-file changes in the PR), and two of the package's own docstrings describe contents/capabilities that contradict what it ships
   - `src/coder_eval/harbor/__init__.py:8`
   - CLAUDE.md's `## Directory Structure` block is the stated architecture reference for this repo, and it gains no `harbor/` entry: `grep -rni harbor CLAUDE.md` returns exactly one line — line 165, where Harbor appears only as "Motivating consumer: an external harness (Harbor / Terminal-Bench 2.0)" inside the execute-vs-run paragraph. So a…
   - Full analysis: `05-architecture.md`
12. 🟡 **Medium [Axis 5]** Thirteen references across seven new modules point at gitignored, untracked `tmp/harborframework*.md` design docs — including an operator-printed warning string, the sole rationale for the `schema_version` wire pin, and a header emitted into every exported Harbor `tests/test.sh` (pinned by a golden fixture)
   - `src/coder_eval/harbor/packager.py:366`
   - `grep -n '^tmp' .gitignore` → `67:tmp/`, `git ls-files tmp/ | wc -l` → `0`, and `ls tmp/harborframework*` → no matches. Yet `grep -rn 'tmp/harborframework' src/` returns 13 references across `harbor/packager.py` (×4), `harbor/experiment_packager.py` (×3), `harbor/agent.py`, `harbor/reward.py`, `harbor/agent_paths.py`,…
   - Full analysis: `05-architecture.md`
13. 🟡 **Medium [Axis 6]** `_write_reference`'s unguarded `shutil.copytree` (packager.py:531) raises bare `FileNotFoundError` past both CLI handlers — leaving a partial export with no `task.toml`, and aborting a whole experiment export the docstring promises it will not abort
   - `src/coder_eval/harbor/packager.py:531` · *cross-axis: 4, 6*
   - `_write_reference` copies unguarded: [code/transcript — see per-axis file] and `_write_environment` does the same at line 281 (`shutil.copy2(source_dockerfile, dest_dockerfile)`). Neither is `TaskNotExportableError`, and `export_command.py:89` catches only `(TaskNotExportableError, CriteriaNotExportableError)`. Reproduced with a task…
   - Full analysis: `06-error-handling.md`
14. 🟡 **Medium [Axis 6]** Over-broad `except ValueError` at run_command.py:1158 wraps the whole batch execution and summary assembly, so an internal/pydantic ValidationError after full agent spend surfaces as `Invalid value: …` exit 2 with the traceback discarded
   - `src/coder_eval/cli/run_command.py:1158` · *cross-axis: 6, 7*
   - The PR wraps the whole run in a handler whose stated target is two guards: [code/transcript — see per-axis file] Those guards are `batch.py:120-131` (`--workspace-dir` requires exactly one task / refuses `driver: docker`) and they are `ValueError` with no dedicated type. But the `try` spans `_run_with_callbacks(...)` → `run_batch(...)`,…
   - Full analysis: `06-error-handling.md`
   - Theme group (1 members): `src/coder_eval/cli/run_command.py:1158`
15. 🟡 **Medium [Axis 7]** `--trajectory` without `--format harbor` is silently discarded: trajectory-dependent criteria score 0.0 and the run reports an ordinary criterion failure
   - `src/coder_eval/cli/evaluate_command.py:326`
   - The forward guard exists — `evaluate_command.py:420-422` refuses `--format harbor` without `--trajectory`: [code/transcript — see per-axis file] and the PR tests it (`tests/test_evaluate_format_harbor.py:116`). The inverse is missing: `trajectory` is declared at :326-332 and read at exactly one place, :424, inside the `if format ==…
   - Full analysis: `07-api-surface.md`

## Nits

1. 🔵 **Low [Axis 1]** Three new module docstrings/comments assert behavior the code does not have (export surface, "registry-derived" coverage, "not built yet")
   - `src/coder_eval/harbor/atif_models.py:22`
   - Full analysis: `01-code-quality.md`
2. 🔵 **Low [Axis 1]** Seven new CLI parameters named `format` shadow the builtin, diverging from the repo's own `report_format` precedent
   - `src/coder_eval/cli/run_command.py:358`
   - Full analysis: `01-code-quality.md`
3. 🔵 **Low [Axis 1]** `_write_environment` calls `_find_workdir` twice on the same freshly-copied Dockerfile, re-reading the file to answer the same question
   - `src/coder_eval/harbor/packager.py:282`
   - Full analysis: `01-code-quality.md`
4. 🔵 **Low [Axis 1]** Two pieces of new public surface do nothing: `export --format` can only ever equal its own default, and `ExportResult.workdir` is read only by tests
   - `src/coder_eval/cli/export_command.py:58`
   - Full analysis: `01-code-quality.md`
5. 🔵 **Low [Axis 2]** `_StepBuilder.append`'s `**step_fields: Any` erases every `Step` field name and type at its six call sites
   - `src/coder_eval/harbor/atif_emit.py:143`
   - Full analysis: `02-type-safety.md`
6. 🔵 **Low [Axis 3]** The nightly Harbor E2E asserts an exact `reward == 1.0` on a score that includes a real `llm_judge` verdict — a float-equality gate on model output
   - `.github/scripts/harbor_e2e.py:114`
   - Full analysis: `03-test-health.md`
7. 🔵 **Low [Axis 3]** The golden `task.toml` hardcodes the full 24-name default `env_passthrough` allowlist twice, duplicating a dedicated test class and coupling the golden to unrelated config churn
   - `tests/_fixtures/harbor_export_golden/expected/task.toml:14`
   - Full analysis: `03-test-health.md`
8. 🔵 **Low [Axis 3]** The test claiming to guard the empty-criteria guard's `to_run` call-site contract asserts a tautology, and its `execute` sibling never checks the exit code
   - `tests/test_detached_grading_guards.py:479`
   - Full analysis: `03-test-health.md`
9. 🔵 **Low [Axis 4]** `docker image inspect` argv takes a task-controlled image with no `--` separator, and trusts its stdout verbatim as the container WORKDIR
   - `src/coder_eval/harbor/packager.py:346`
   - Full analysis: `04-security.md`
   - CVSS: `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:L/A:N`
10. 🔵 **Low [Axis 6]** `harbor/agent.py`'s two import handlers: one is dead code that would silently substitute a fake version, the other reports "harbor not installed" for any deeper ImportError from an installed harbor
   - `src/coder_eval/harbor/agent.py:65`
   - Full analysis: `06-error-handling.md`
11. 🔵 **Low [Axis 6]** The Harbor E2E script's blanket `except Exception` prints only `str(exc)` with no traceback, so the nightly's only diagnostic for an unexpected failure is a bare message
   - `.github/scripts/harbor_e2e.py:148`
   - Full analysis: `06-error-handling.md`
12. 🔵 **Low [Axis 7]** `--workspace-dir` is documented as an absolute path in three places with nothing enforcing it, and the underlying docstring still calls it DockerRunner-internal
   - `src/coder_eval/cli/execute_command.py:195`
   - Full analysis: `07-api-surface.md`
13. 🔵 **Low [Axis 7]** `coder-eval export` merges into an existing output directory with no `--force`/`--clean`, so a stale variant or reference file survives a re-export
   - `src/coder_eval/cli/export_command.py:43`
   - Full analysis: `07-api-surface.md`
14. 🔵 **Low [Axis 7]** `portability.py` hard-errors NEEDS_TRAJECTORY criteria citing a capability this same PR ships
   - `src/coder_eval/harbor/portability.py:23`
   - Full analysis: `07-api-surface.md`

## What's Missing

- 🟠 **parallel-paths** — The deleted model invariant was replaced at ONE call site (`cli/run_command.py:1140`, fed `to_run`) — the three other paths that consume a `TaskDefinition` were not updated: `run_evaluation` (`cli/evaluate_command.py`), `run --resume`'s `to_grade` set (`orchestration/regrade.py::regrade_in_place`), and `coder-eval plan`. Verified: `grep…
- 🟠 **parallel-paths** — The Harbor verifier is a THIRD grading environment (Harbor's own container, tempdir driver — `_write_verifier_task_yaml` emits no `sandbox:` block at all, so the task's provisioning, `env_passthrough` and `working_dir` are absent, and `pre_run` is never run) yet nothing stamps the row the way `graded_on_host` / `graded_without_pre_run`…
- 🟡 **parallel-paths** — `shutil.copytree` at `packager.py:449` and `:531` are the only reference/template copies in `src/` that omit `ignore=ignore_patterns_and_symlinks(REFERENCE_COPY_IGNORE)`; the four sibling sites (`orchestration/evaluation.py:125`, `isolation/docker_runner.py:1295` and `:1322`, `evaluation/sub_agent.py:154`) all pass it. A new copy site…
- 🟡 **parallel-paths** — `packager.py:527` re-implements the reference-path join (`task_file.parent / task.reference.directory`) instead of calling `orchestration/evaluation.py:30::resolve_host_reference_dir`, whose own docstring calls itself "the ONE place that knows how a reference path is spelled relative to its task file" — a second spelling of the same…
- 🟡 **parallel-paths** — The new `harbor = ["harbor==0.22.0"]` extra was wired into both pyright jobs (`pr-checks.yml:85`, `:392`) but into no contributor-facing install path: `make install` is still `uv sync --frozen --extra dev --extra uipath` (labelled "parity with CI"), and `git diff --name-only origin/main...HEAD` shows no `Makefile`, `CONTRIBUTING.md` or…
- 🟡 **parallel-paths** — `--format` was added to four commands with three independent validators and two exit codes (`run_command.py:472` BadParameter/exit 2, `evaluate_command.py:410` Exit(1), `export_command.py:31`+`:79` its own `_SUPPORTED_FORMATS` tuple + a separate default literal at `:59`) — the allowlist was modelled generically in exactly one of the…
- 🟠 **tests** — Neither new command is ever invoked through `CliRunner` (`grep -rn export_command tests/` returns nothing): `export_command.py` is at 17.54% (lines 79-123 — every refusal, the experiment branch and the all-skipped exit gate) and `harbor_command.py` at 57.14% (lines 57-62 — the `RewardWriteSkippedError`/`RegradeError` → exit-1 mapping…
- 🟠 **tests** — The contract binding the PR's two halves — the flat promotion at `run_command.py:692-695` and the flat read at `harbor/agent.py:139` — has no shared assertion: deleting the promotion leaves the suite green (verified: 48 passed with lines 692-695 removed), and the `len(written) != 1` arm is never executed. The only safety net is…
- 🟡 **tests** — `harbor` is now an installed CI extra, yet no test ever runs against the real package: `tests/test_harbor_agent.py:28-45` monkeypatches a FAKE `BaseInstalledAgent` into `sys.modules`, so `CoderEvalAgent`'s conformance to Harbor's real ABC is only ever checked statically by pyright, and `tests/test_atif_models.py:5-14` still documents a…
- 🟡 **tests** — No test exports a dataset-backed task on either path: the single-task path has no `dataset` guard at all, and `experiment_packager.py:122` (`dest = dest / resolved.task.row_id`, the documented `<row_id>` segment) is never executed — which also means nothing catches that `task_loader.py:464` already rewrites an expanded row-task's…
- 🟡 **tests** — The 4 E2E scenarios (`harbor_e2e.py:42-55`) are all happy-path single-task exports: none passes `-e <experiment.yaml>` (which is why the agent-override skip that makes every `-e experiments/default.yaml` export emit zero directories shipped unnoticed), none exercises the load-bearing "write NOTHING" reward contract (an ungraded/crashed…
- 🟡 **tests** — `grep -rn model_fields tests/test_atif_emit.py tests/test_atif_hydrate.py tests/test_atif_models.py` returns nothing: the emit→hydrate round trip is asserted only against a hand-picked subset, which is how `CommandTelemetry.result_status` and `duration_ms` came to be emitted and never read back. The repo already has the fail-closed…
- 🟠 **downstream-consumers** — `compute_reward` publishes `result.weighted_score` and consults no gate, so coder-eval's own verdict logic has no downstream representation in Harbor: a row whose criterion sits below its `pass_threshold` (FAILURE / exit 1 natively) still publishes e.g. `reward: 0.95`; `all_criteria_passed` / `armed_criteria_passed` (the FIRED-ONLY…
- 🟠 **downstream-consumers** — `run_limits` has exactly one mapped key (`task_timeout` → `[agent]`/`[verifier] timeout_sec`, `packager.py:539-555`); every other consumer of that namespace is silently unrepresented, so the exported agent phase falls back to the packaged default experiment's caps (`max_turns 100`, `turn_timeout 300`) and runs with NO `max_usd` / token…
- 🟡 **downstream-consumers** — The new published aggregate drops one of the four token buckets: `atif_emit.py:348` sets `total_cached_tokens = total.cache_read_input_tokens` only, so cache-CREATION tokens (which CLAUDE.md calls a dominant cost driver) appear in no `FinalMetrics` field, and `harbor/agent.py:155-157` feeds that straight into `context.n_input_tokens` /…
- 🟡 **downstream-consumers** — The removed invariant is still asserted as fact in the user-facing guide — `docs/TASK_DEFINITION_GUIDE.md:702` ("Every task needs at least one success criterion") — and no doc surface was touched by this PR. CE030's doc/schema parity rule tracks FIELDS, not invariants, so removing a validator is invisible to `make lint`.
- 🟡 **display-mapping** — `_PORTABILITY_BY_TYPE` (`portability.py:57-72`) classifies by criterion TYPE only, so instance-level trajectory dependence is unclassified: `llm_judge` / `agent_judge` are `NEEDS_CREDENTIALS` alone, yet their `include_agent_output` / `include_tool_calls` / `include_dialog` options (`models/criteria.py:1172-1198`) read the `iterations`…
- 🟡 **display-mapping** — The new run-dir artifact is in no filename registry and in no layout contract: no `TRAJECTORY_JSON_FILENAME` in `path_utils` (so CE053, which derives its target set from those constants, is structurally blind to it), and neither `<task>/trajectory.json` nor the promoted `<run_dir>/trajectory.json` appears in…
- 🔵 **display-mapping** — Two commands were registered (`cli/__init__.py:87-89`) and added to no command-listing surface: `grep -rln "coder-eval export" docs/ README.md plugins/ .claude/` returns nothing, so `docs/USER_GUIDE.md`'s per-command reference (which has a section for every other command, incl. `aggregate`), `mkdocs.yml`'s `nav:`/`extra.docs_index`, and…
- 🟡 **daily-nightly** — A new paid nightly is introduced with no statement of its cost, ownership or failure handling: `harbor-e2e.yml` runs on `schedule` (05:17 daily) AND `push: branches: [main]`, doing two `docker build`s and real Bedrock calls (incl. a verifier-phase `llm_judge`) per run, with a 20-minute budget. It uploads no artifacts on failure (every…
- 🟡 **daily-nightly** — The PR changes the shared production run path, not just the Harbor path: the new `try` at `run_command.py:1146-1159` now wraps `run_batch` and the post-run `RunSummary` assembly, so a nightly whose `_check_task_count_invariant` (a `ValueError`, and `pydantic_core.ValidationError` subclasses `ValueError`) trips is reported to the…
- 🟡 **daily-nightly** — Three cross-repo contract surfaces gained new shapes with no statement of downstream impact: a new per-task run-dir artifact plus a run-root `trajectory.json`, and two new public flags (`--format`, `--workspace-dir`) on `run`/`execute`. `docs/REPORT_SCHEMA.md` (the stated run-record/report contract for the external eval-runner +…
- 🔵 **daily-nightly** — `harbor_e2e.py:51` makes the in-tree CI task `tasks/byod_smoke_test.yaml` a fixture for a second, cross-workflow pipeline, and nothing in that task file says so — adding a non-portable criterion to it for coder-eval's own CI (e.g. `cli_called`, which `portability.py` hard-refuses) would break the Harbor nightly with an error naming the…

**Static checks (lint / type)**

- *(ruff)* Add `A` (flake8-builtins) to `[tool.ruff.lint] select` in pyproject.toml (currently `E,F,I,N,W,UP,B,SIM,RUF,PLR0915,PLR0912`), with a `per-file-ignores` entry only if a real builtin-named field is unavoidable. A002 fires on a function argument shadowing a builtin. — prevents: A1-low: seven new parameters named `format` (cli/run_command.py:358, :461, :575; cli/evaluate_command.py:316, :381;…
- *(pyright)* Replace the free-string `--format` with a closed type: one `InterchangeFormat` StrEnum in a cycle-free `models/` leaf (or `harbor/`), typed `InterchangeFormat | None` on `run_pipeline` / `run_evaluation` / `export_command` and declared to Typer as an enum (Click renders the choice list and refuses… — prevents: A1/A2/A7-medium: six `"harbor"` literals across three commands (run_command.py:472, :686; evaluate_command.py:410, :413;…
- *(pyright)* Make the local type gate equal the CI one: change `make install` (Makefile:15) from `uv sync --frozen --extra dev --extra uipath` to also install `--extra harbor --extra codex --extra litellm`, and add a derived lint test asserting the Makefile's sync extras are a superset of the extras used by… — prevents: A2-medium: `uv run pyright` reports 6 errors on a clean checkout (3 new — harbor/agent.py:52, :53, :56 — after commit c2b79d7d…
- *(ce-lint)* CE058 — every `shutil.copytree` / `shutil.copy2` of a task-authored tree in `src/` must pass an explicit `ignore=`, and a reference/template/plugin tree must use `ignore_patterns_and_symlinks(...)` from `path_utils` (symlinks dereferenced by default is the hazard). Wire as a new AST rule in… — prevents: A4-medium: harbor/packager.py:531 (`tests/reference/`) and :449 (`environment/templates/`, then baked into the image by the…
- *(ce-lint)* CE059 — a `subprocess.run` / `create_subprocess_exec` argv list must place a `"--"` literal before any trailing non-literal positional (a value read off a task model or a run record). Same AST shape as the existing CE015 argv rule; wire into `tests/lint/rules/` + `runner.py`. — prevents: A4-low: harbor/packager.py:345-352 runs `["docker", "image", "inspect", image, "--format", ...]` with an unvalidated task-YAML…
- *(ce-lint)* CE060 — no `.format()` / f-string interpolation of an unquoted value into a string that is later written to a `.sh` file or a Dockerfile (detect by the destination path suffix on the `write_text` call, or by a `#!/bin/sh` prefix in the template literal); require `shlex.quote(...)` on every… — prevents: A4-high (CVSS AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H): `_TEST_SH_TEMPLATE` (harbor/packager.py:87) interpolates…
- *(ce-lint)* CE061 — no `tmp/` path in any string literal, docstring or emitted template under `src/`. `tmp/` is gitignored (.gitignore:67, `git ls-files tmp/` = 0), so every such reference is dangling by construction. Add the accompanying fix as a tracked `docs/HARBOR_INTEROP.md` (mkdocs `nav:` +… — prevents: A5-medium: 13 references to the untracked `tmp/harborframework*.md` across 7 new modules, including (a) an operator-printed…
- *(ce-lint)* CE062 — widen CE053 from a constant-derived allowlist to a PATTERN detector: any bare string literal matching `^[a-z0-9_.]+\.(json|log)$` used in a `Path` join / `.with_name()` / `/` expression under `src/` must resolve to a `path_utils` constant. Today… — prevents: A5-medium: `trajectory.json` ships as three bare literals across three packages — producer harbor/atif_emit.py:418 (in a…
- *(ce-lint)* CE063 — ATIF wire-key parity: every `extra[...]` key literal written in `harbor/atif_emit.py` (`cache_creation_tokens`, `reasoning_tokens`, `message_id`, `stop_reason`, `result_status`, `duration_ms`, `crashed`, `crash_reason`, `iteration`, `reconciliation`) must either be read back in… — prevents: A8-high: `atif_hydrate` reads NONE of them (`grep -c extra` in that file returns 0), so `CommandTelemetry.result_status` and…
- *(ce-lint)* CE064 — no bare `weighted_score is None` (or `weighted_score == 0.0`) measured/unmeasured test outside `models/results.py`; route every such gate through one predicate (`nothing_was_measured`, or a new `EvaluationResult.is_measured` that also treats `final_status.category == "error"` as… — prevents: A6-high: harbor/reward.py:65 is exactly `if result.weighted_score is None:`, but `None` is written only on…
- *(ce-lint)* CE065 — no `**kwargs: Any` splatted into a pydantic `BaseModel(...)` constructor; require a `TypedDict(total=False)` + `Unpack[...]`, or pass the constructed model. Family sibling of CE050 (untyped union `getattr` probe) — both forbid a type-erasing hop that makes pyright blind to a model's field… — prevents: A2-low: `_StepBuilder.append(self, parent_tool_use_id, **step_fields: Any)` (harbor/atif_emit.py:143) is the ONLY boundary…
- *(ce-lint)* CE066 — architecture/onboarding surface parity (derived test, `@pytest.mark.lint`, same shape as `SKILL_DOC_SURFACES` in tests/test_custom_lint.py:1397 and CE047): every package directory under `src/coder_eval/` and every `cli/*_command.py` must be named in CLAUDE.md's `## Directory Structure`… — prevents: A5-medium: `grep -rni harbor CLAUDE.md` returns ONE line (165, where Harbor is only named as a competitor/consumer), and `git…
- *(ce-lint)* CE067 — CLI test parity (derived test): every non-hidden command registered on the Typer app must be invoked at least once through `CliRunner` somewhere in `tests/`. Enumerate from `cli/__init__.py`'s `app.command(name=...)` / `add_typer` registrations and grep `tests/` for the command token in a… — prevents: A3-high: `grep -rn "export_command" tests/` returns NOTHING — `cli/export_command.py` is at 17.54% with its entire body (79-123)…
- *(ce-lint)* CE068 — vendored-mirror version pin (derived test, `tests/test_action_version_pin.py` precedent): a module that declares itself a mirror of a third-party schema must name the version pinned in `pyproject.toml`; fail on drift in either direction. — prevents: A2-high (the detectable half): `harbor/atif_models.py:3` says it mirrors "harbor 0.20.0" and `tests/test_atif_models.py:5` says…
- *(ce-lint)* CE069 — complexity ratchet (`radon` is ALREADY a runtime dependency, pyproject.toml:48, so this costs nothing to add): a `@pytest.mark.lint` test that runs `radon cc` over `src/coder_eval/` and fails on (a) any function above CC 20, and (b) any function in the 10-20 band that is not in a frozen… — prevents: A1-medium: four new functions land in the 10-20 band (`harbor/atif_emit.py:172 _emit_turn` C(19), `harbor/atif_hydrate.py:100`…
- *(ce-lint)* CE070 (lower confidence — measure FP volume before wiring) — dead public surface: a public module-level function or model field in `src/` whose only references in `src/` are inside its own defining module, or whose only readers live in `tests/`, must be made private, inlined, or exported in… — prevents: A1-medium/low pair: `write_trajectory_json_strict` (harbor/atif_emit.py:366) is public, ungated by any `__all__`, called from…

**Harness improvements**

- Add an experiment scenario to `.github/scripts/harbor_e2e.py`: `export_task` (:66) builds `["coder-eval","export",task_file,"-o",out_dir]` and no scenario in `SCENARIOS` (:42-55) ever passes `-e/--experiment`, so the entire `export_experiment` path — variants, replicates, lineage-based skip…
- Make the Harbor round trip a `pull_request` check (a trimmed, credential-free subset is enough — `.github/workflows/harbor-e2e.yml` currently triggers only on `workflow_dispatch`, nightly `schedule`, and `push: branches: [main]`).
- Restore the empty-criteria invariant at a layer every graded entry point crosses — keep the `TaskDefinition.success_criteria` validator this PR deleted and give the Harbor agent-phase YAML an explicit opt-out field (so `packager.py:510`'s `"success_criteria": []` is the only thing that can express…
- Add an ATIF round-trip parity test: build one fully-populated `EvaluationResult`, run `evaluation_result_to_trajectory` → `trajectory_to_turn_records`, and assert a fail-closed MAPPED / DELIBERATELY-DROPPED partition over `set(TurnRecord.model_fields)`, `set(AssistantMessage.model_fields)` and…
- Replace the manual "reproducible procedure" comment in `tests/test_atif_models.py:5-14` with a real parity test guarded by `pytest.importorskip("harbor.models.trajectories")`, walking `model_fields` (name, annotation, required, metadata) for all 10 mirrored models and asserting superset-on-read /…
- Add an export completeness + round-trip test: run the exporter, then (a) assert no `${row.` residue survives in the emitted `environment/task.yaml` or `tests/task.yaml`, (b) re-load both emitted YAMLs through `TaskDefinition` and assert they validate, and (c) assert a fail-closed partition of…
- Close the `--workspace-dir` staleness gap: narrow the stale-artifact warning suppression at `orchestrator.py:1657` from `self.workspace_dir is None` to the in-container case it was written for (gate on `IN_CONTAINER_ENV` from `models/container_paths.py`, per CE056, NOT on the field), and add a…
- Add a work-dir immutability + idempotence test for `evaluate`: after `evaluate <task.yaml> <workdir> --format harbor --trajectory t.json`, assert the work dir gains no run-record file, that no "original kept as task.execute.json" notice is printed, and that the identical invocation succeeds a…
- Add a coverage ratchet for new files: the global `--cov-fail-under=80` cannot see a 17%-covered new module, so record per-file coverage in a baseline and fail when a file lands (or drops) below its entry. Pair it with the diagnostics fixes in `.github/scripts/harbor_e2e.py`: add…
- Make the export golden fixture insensitive to unrelated config: mask the two `*.env` tables in `_relative_files` before diffing, or narrow the golden source task's `env_passthrough` to an explicit two-name list. `tests/_fixtures/harbor_export_golden/expected/task.toml` embeds the full 24-name…
- Add a CLI flag-interaction matrix test for the new `--format` / `--trajectory` / `--workspace-dir` / `--junit-xml` / `--allow-host-grading` pairs: for every pair where one flag's refusal names the other, assert BOTH directions refuse, with one consistent exit code. Then add the missing inverse…

## Top 5 Priority Actions

1. Restore the empty-`success_criteria` refusal at a layer every graded path crosses: the model validator was deleted (src/coder_eval/models/tasks.py:495) and its replacement guard has one call site fed only `to_run` (src/coder_eval/cli/run_command.py:1140), so `coder-eval evaluate` and `run --resume`'s `to_grade` set both grade a criteria-less task as SUCCESS at weighted_score 0.000 and exit 0 — add the check to `orchestration/regrade.py::regrade_in_place` plus `run_evaluation`, and extend the CLI call to `part.to_grade`.
2. Widen the unmeasured-row guard in `write_reward` from `weighted_score is None` to also refuse `final_status.category == "error"` (src/coder_eval/harbor/reward.py:65), because a grading pass that crashes on `JudgeInfrastructureError`/`CheckerMisuseError`/`ReferenceTamperedError` finalizes ERROR with empty results and a hard `weighted_score = 0.0` (src/coder_eval/models/results.py:661), which the verifier's `|| true` then publishes to Harbor as a measured `reward: 0.0` — training 'the agent behaved badly' from a measurement that never happened.
3. Read `result_status` and `duration_ms` back out of `ToolCall.extra`/`ObservationResult.extra` in `_tool_call_to_command` (src/coder_eval/harbor/atif_hydrate.py:48) and reconcile which ATIF slot carries `result_status` (emit puts it on the tool call, tests/fixtures/atif/known_good_trajectory.json:36 puts it on the observation), since today `command_executed(require_success: true)` scores 1.0 natively and 0.0 through `evaluate --format harbor` for the same trajectory.
4. Replace the two hand-picked export payloads with an allowlist validated against `set(TaskDefinition.model_fields)` so an uncarried field fails the export loudly (src/coder_eval/harbor/packager.py:503 and :398), carrying `run_limits` and `checker_context` through and refusing `simulation.enabled` and `dataset:` tasks beside the existing driver refusal at src/coder_eval/harbor/packager.py:166 — today declared turn/USD/token budgets are silently replaced by baseline defaults, a pinned judge route is lost, and a dataset task exports one directory whose prompt and criteria still contain literal `${row.*}` placeholders at exit 0.
5. Make the newly host-reachable `--workspace-dir` honest about staleness by narrowing the stale-artifact warning suppression at src/coder_eval/orchestrator.py:1657 from `self.workspace_dir is None` to the in-container case it was written for (gate on `IN_CONTAINER_ENV` per CE056, not on the field) and documenting that `clear_rerun_artifacts` deliberately excludes the workspace — verified live, a pre-seeded `stale.txt` gave a no-op agent `file_exists` 1/1 and score 1.000 with no warning; then fix the three feature-blocking/ hardening items in the same pass: delete the falsified agent arm of `_unhonorable_override_reason` (src/coder_eval/harbor/experiment_packager.py:84) that makes every `export -e` emit zero directories and exit 1, `shlex.quote` the workdir interpolated into the generated `tests/test.sh` (src/coder_eval/harbor/packager.py:87), and pass `ignore=ignore_patterns_and_symlinks(REFERENCE_COPY_IGNORE)` at src/coder_eval/harbor/packager.py:531 and :449 so an exported bundle stops inlining host secrets.

---

**Adversarial verification:** 63 findings proposed → 46 verified, **6 refuted as false-positive and dropped**, 34 corrected in place; 17 🔵 low findings passed through unverified. 2 further survivor(s) dropped by cross-axis reconcile.

<sub>8-axis structured review · deterministic scoring `10 − 3.0·🔴 − 1.0·🟠 − 0.5·🟡 − 0.1·🔵` · every medium+ finding independently re-checked by a second agent instructed to refute it.</sub>

uipreliga:
Let's see if they like it... i thought that the easiest approach would have been to simplify the tasks, remove some of the criteria while move others to the testing scripts. If some tasks are not possible to score in Harbor, we shouldn't just give them out.

uipreliga (approved):
🚢

PR #175, opened 2026-09-15
feat(harbor): bind-mount task.yaml/plugins/templates/extra_mounts, skip Dockerfile when unneeded, translate pre_run

akshaylive:
## Summary

- `environment/docker-compose.yaml` now bind-mounts `environment/task.yaml` itself, each `type: local` `agent.plugins[]` dir, each `TemplateDirSource` in `sandbox.template_sources`, and each `sandbox.docker.extra_mounts` entry (its own `ro`/`rw` mode preserved) — all at their own host path, instead of `COPY`ing content into the image and rewriting paths in `task.yaml`. Fixes a real bug: an exported agent previously ran with no skill/plugin content and no credentials (e.g. `~/.uipath`) available inside the container, since `docker_runner.py`'s auto-mount for a normal docker sandbox run had no Harbor equivalent. Mount paths are always emitted POSIX-style (`as_posix()`), since docker-compose volume specs are POSIX regardless of the exporter's host OS.
- `environment/Dockerfile` is now written only when `sandbox.docker.dockerfile_path` is set (real `RUN` build steps needed); otherwise `task.toml`'s `[environment].docker_image` points Harbor straight at the pre-built image via its own `should_use_prebuilt_docker_image` path, and no Dockerfile is written at all.
- `pre_run` commands are now translated into the agent-phase `task.yaml` — they run before the agent starts, which `coder-eval execute` still does for the `CoderEvalAgent` embed, so this was a fixable gap rather than a fundamental limitation (unlike `post_run`, which belongs to the grading phase `execute` never runs, and stays untranslated with a warning).

## Test plan

- [x] `pytest tests/test_harbor_packager.py tests/test_harbor_export_golden.py tests/test_harbor_experiment_packager.py tests/test_harbor_agent.py` — 59/59 passing
- [x] Full suite: `pytest` — 6167 passed, 8 skipped
- [x] Exported and ran `uipath-admin` audit smoke task end-to-end through `harbor run -a coder_eval.harbor.agent:CoderEvalAgent -e docker`, reward `1.0`
- [x] Exported and ran `uipath-agents antipattern_openai_agents_hitl` (which needs `pre_run` to seed its fixture) end-to-end, reward `1.0`


uipreliga (approved):
🚢

PR #177, opened 2026-09-15
docs: move design rationale out of src/ into .claude/notes, gate it, and slim CLAUDE.md

uipreliga:
Cuts 79,754 words of essay-shaped prose in `src/coder_eval` to 17,752, and 126 docstrings over 150 words to **zero**, by moving design rationale into `.claude/notes/` and leaving a resolvable pointer at each source site.

**No code changed.** `python -m tests.lint.prose_budget --assert-code-unchanged 9af3db3d` compares a docstring-stripped `ast.dump` plus a per-file multiset of functional directive comments (`# noqa`, `# type: ignore`, `# pyright:`, `# nosec`, `# pragma:`, `# fmt:`) and reports nothing across all 14 commits. All 116 directives intact.

## What moved, and what stayed

Every passage was classified: CONTRACT and HAZARD stay in source (compressed); RATIONALE moves to `.claude/notes/<subsystem>.md`; HISTORY is deleted (git holds it); a CROSS-MODULE CLAIM is deleted and linked to its owner. `.claude/architecture-notes.md` is replaced by nine subsystem files under one index.

## The gate

`make docs-budget`, wired into both `make verify` and `pr-checks.yml`'s `quality-gate` (CI restates every step as `.venv/bin/...` and never calls `make`, so both seams were needed). Four checks:

- every `Rationale: <path> § <heading>` pointer resolves
- a pointer is the LAST line of its block — catches a rewrite that strands the tail of the replaced prose after it, where the `$`-anchored regex cannot see it
- own-line comments per file ≤ `MAX(20, 0.15 × file length)`
- no docstring over 150 words of PROSE

The last two replaced a hand-maintained tree-wide word count. Nothing in the gate is now a number anyone maintains: delete code and the comment budget shrinks with it, and a new file is governed from its first commit. Exemptions are categorical — an `Args:`/`Returns:`/`Raises:` block is structure not prose, and an `@abstractmethod` docstring IS the interface contract implementers read.

## Reviewing this

The diff is large but uniform. The highest-value thing to check is **whether any compressed docstring now asserts something false** — that produced every real defect found here, and the worst case was a contract not dropped but *replaced* with a plausible wrong reason that invited the regression it originally forbade. Those are fixed in `0fe3e0a7`, `fd72e9aa` and `f7b5a193`; the commit messages name each one.

Three known items, all recorded rather than hidden:
- `isolation.md` still carries a 19KB single-line bullet from Phase 1. A reviewer wanted it deleted as duplicated; measurement says later phases paraphrased rather than copied it, so deleting would lose detail. Left as a judgement call.
- `tests/` (105,417 lines) is unmeasured — the plan's own open question.
- Four harness gaps are deferred with reasons in `.claude/harness-candidates.md`; the strongest is a single-home detector for the notes tree.

## Verification

5,733 tests / 2 skipped · `make lint` 635 · pyright 0 · coverage 92.60% · `make evalboard-verify` green · `make plugin-reference` and `make docs-indexes` produce no diff · all 8 public `--help` outputs byte-identical to before the work.

---

## Added after review

Review fixes, then a second pass over `CLAUDE.md` itself.

**From the review** — the prose gate penalized usage examples: `_TRAILING_SECTIONS` accepted an `Example:` block after a pointer while `_DOCSTRING_SECTIONS` counted it against the 150-word bar, so a docstring in the house style could fail the gate. Both deleted examples are restored, with `execute_with_retry`'s keyword contract. The CE catalogue in `notes/README.md` (2,165 words, disclaimed as non-authoritative, targeted by no pointer) and "Not wired up yet" in `permissions.md` (769 words designing code that does not exist) are deleted. `_TYPER_COMMANDS` gains a staleness test.

**Not done, with evidence** — the review also asked for the per-file abstracts to go as restatements of the sections beneath them. 61 of the isolation abstract's 191 code identifiers appear nowhere else in `notes/`, `docs/` or `CLAUDE.md`, including `--allow-recorded-commands`, `task_config.resolved` and `_seed_from_prior_result`'s `early_stop` carry. Deleting them loses rationale rather than a duplicate. The readability half is fixed instead: every wall bullet is reflowed to 88 columns (the largest was a single 19,453-character line), and `reporting.md`'s pseudo-tree becomes prose. Re-homing those claims is a follow-up.

**`CLAUDE.md`: 2,802 → 2,028 words.** It is loaded every session, so a copy kept there costs context on every turn and goes stale with nothing to sense it. Three sections were copies:

- **Directory Structure** was wrong — it listed `optimize/`, which exists on neither this branch nor main, and omitted `errors/` and `plugins.py`, both present since the initial public release. Now a list of what a filename cannot tell you, opening with "run `ls`".
- **Success Criteria** was a third copy of one registry, behind the CE033-generated plugin reference and the Task Definition Guide, with no parity test — the shape CE030's own `CLAUDE.md` test exists for.
- **Extension Points** restated `docs/EXTENDING.md` step for step, 190 lines below a section that says "Each entry is a pointer".

Also reverts a run-limit parity sentence claiming an adapter rejects an unsupported field at load time. It does not: `opencode_agent.py:785` and `pi_agent.py:738` both warn and continue. Neither lint-pinned surface is touched.

**Rebased onto `7bf89661`** (the harbor bind-mount rewrite). Five conflicts in `harbor/packager.py`, all resolved as main's code plus this branch's prose rules; two of its new docstrings were over the bar and are slimmed, with the rationale moved into `reporting.md`.

Verification after the rebase: 5,737 tests / 2 skipped · `make lint` 635 · pyright 0 · coverage 92.57% · `docs-budget` 0 essays · all 15 CI checks green.


bai-uipath (approved):
**Fix what you agree with, otherwise lgtm.** The prose-only invariant holds: docstring-stripped ASTs are identical across all 99 changed Python files, directive comments are unchanged, Typer help text is byte-identical, and the judge's verdict schema is untouched. The pointer convention is real rather than decorative, with 363 pointers in src resolving to 83% of the note headings and none dangling.

Worth stating before the cuts: src prose drops from 77,954 words to 17,752, but the notes go from 9,356 to 41,391, so the new home is 4.4x the old one. The cuts below take roughly a quarter off that.

## Cuts

- **The CE lint-rule catalogue in the notes README.** Around 2,165 words duplicating each rule's own module docstring, disclaimed as non-authoritative in the two sentences directly above it, and targeted by no pointer. Fix: delete the catalogue, keep the sentences that name where the authority lives.
- **The per-file abstract lines.** Seven single-line summaries, the largest 19,000 characters, restating the sections immediately beneath them: about 6,900 words, 17% of the notes. Fix: delete them, since the contents list and the headings already navigate.
- **"Not wired up yet" in the permissions note.** 769 words, nearly half that file, designing code that does not exist. Fix: open an issue and delete the section.
- **The CLAUDE.md commit.** It asserts that an adapter rejects an unsupported run-limit field at load time; the opencode and pi adapters both log a warning and continue, and the parity doc still records those divergences as table rows. Fix: revert that sentence. The communication-style and delete-before-you-guard edits in the same commit aren't about prose relocation either.

## Worth fixing

- **The gate penalizes usage examples.** Args/Returns/Raises blocks are excluded from the word count but Example: blocks are not, so a call example counts against the bar. Two were deleted here, including the keyword contract for the retry helper. Fix: add Example:/Examples: to the excluded sections and restore them.

Minor: the Typer exemption set has no stale-membership test, unlike the other allowlists in this repo; and the abstract-method exemption is carrying one 315-word docstring worth trimming on its own merits.

uipreliga:
Thanks — three of the four cuts and the gate fix were right. One cut I could not verify, and I explain below why I did not apply it. Changes are staged locally and land in the next push.

## Applied

**The gate penalizes usage examples.** Correct, and the inconsistency was internal: `_TRAILING_SECTIONS` already accepts an `Example:` block after a pointer, while `_DOCSTRING_SECTIONS` counted it against the 150-word bar. The two lists disagreed about the same shape. Added `Example:`/`Examples:` to the exclusion, restored both deleted examples plus the `on_attempt_error` keyword contract, and added two tests — one for the exclusion, one asserting the two lists agree on `Example:`.

**The CE lint-rule catalogue.** Deleted. Nothing pointed at it, and the two sentences above it already disclaimed it. `README.md` drops 2,745 → 589 words.

**"Not wired up yet".** Deleted from `permissions.md`.

**The CLAUDE.md parity sentence.** Reverted. Confirmed: `opencode_agent.py:785` and `pi_agent.py:738` both collect `_UNSUPPORTED_CONFIG_FIELDS`, log a warning, and continue. Nothing rejects at load time.

**Minor.** Added the stale-membership test for `_TYPER_COMMANDS` — each pair must name a file that exists and a function defined at module level there, the same shape as CE057's. Trimmed `Agent.communicate` from 315 to 273 words.

## Not applied: the per-file abstract lines

The premise does not hold. I checked the isolation abstract's 191 code identifiers against the rest of its own file, every other note, all of `docs/`, and `CLAUDE.md`. **61 appear nowhere else.** Among them:

- `--allow-recorded-commands` — the untrusted-recorded-config trust gate
- `task_config.resolved` — why run-dir mode does not re-load the YAML
- `_seed_from_prior_result`, `Orchestrator._select_gate()`, and the `early_stop` carry that makes fired-only gate selection correct
- `FinalStatus.is_execution_fact` and the `MAX_TURNS_EXHAUSTED` exclusion
- `CONTAINER_GRADE_WORKSPACE` / `CONTAINER_OUTPUT_DIR`, `_sanitize_restored_path`, `O_EXCL|O_NOFOLLOW`, `_operator_baseline_post_run()`

The smaller abstracts behave the same way. The `contracts.md` dataset bullet is the only place in `notes/` that records the sampling-determinism contract. So these lines are not restating the sections beneath them — they sit above material the sections never reach. Deleting them loses rationale rather than a duplicate.

The readability complaint is fair, so I fixed that part without deleting anything:

- Reflowed every wall bullet to the notes' 88-column prose width. The isolation one was a single 19,453-character line.
- Converted `reporting.md`'s pseudo-directory-tree — two entries carrying 2,174- and 506-character trailing comments — into `###` subsections with ordinary prose.

If the abstracts should still go, the safe order is to re-home those 61 claims into the `###` sections first, then delete. Happy to do that pass if you want it in this PR.

## Not applied: the rest of the CLAUDE.md commit

The communication-style move and `delete before you guard` are off-topic for a prose-relocation PR, agreed. Both statements are accurate and `delete before you guard` is the principle the notes README argues from, so reverting them is churn without a correctness gain. Say the word if you would rather they went to their own commit.

`make verify` is green: 5736 tests, 635 lint, pyright clean, `docs-budget` reports 0 essays.

PR #106, opened 2026-08-11
feat(reference)!: directory-only references + anti-cheat permission window

akshaylive:
Reference solutions were readable by the agent under evaluation — it shares a filesystem with the harness, so an agent could grep the task directory for the answer instead of solving the task. This closes the passive-read path and adds an adversarial probe to CI.

## Read this first: known gap

**This is defense-in-depth, not a boundary.** `chmod(2)` is gated on *owner-or-`CAP_FOWNER`*, and the container runs as **root**, which owns the bind-mounted reference copy. An agent that deliberately runs `chmod 755 /work/references` regains access:

```
orchestrator chmod 000 -> rc=0
  agent read while shielded: Permission denied
  AGENT RUNS chmod 755      : rc=0 (SUCCEEDED)
  agent read after chmod    : CANARY=42
```

Verified on Docker Desktop **even with `DAC_OVERRIDE`, `DAC_READ_SEARCH`, `FOWNER` and `CHOWN` all dropped** — the owner check passes with no capability at all. Dropping those caps does close it where the mount preserves a non-root owner (native Linux), so they are kept.

Passive reads are blocked; an adversarial agent is not. **Full containment requires running the agent as a non-root uid that does not own the reference** — deliberately left as follow-up rather than bolted on here. Documented in `CLAUDE.md`, `docs/DOCKER_ISOLATION.md`, `docs/TASK_DEFINITION_GUIDE.md`, the module docstring, and the probe task itself, so nobody reads a green probe as "the reference is unreachable".

## What changed

- **`task.reference` is directory-only.** `code:`/`file:` removed — a directory is the only shape that can be permission-gated as a unit. The removed keys raise a migration error naming the replacement. `reference_comparison` gains a required `reference_file`.
- **`orchestration/permissions.py::set_permissions`** — async CM that chmods paths for a block. Windows **stack**, so a nested re-grant (`mode=READ_ONLY_MODE`) restores the *enclosing* mode, not the original; the stack also subsumes what a refcount would do. Crash-safe: `finally` + `asyncio.shield` on both halves + `atexit` + chained SIGINT/SIGTERM.
- **`Sandbox.set_permissions`** enforces only inside a container, gated on `CODER_EVAL_IN_CONTAINER` — **not** on `sandbox.driver`, which the in-container entrypoint rewrites to `tempdir` (a driver-based gate would silently disable the feature on exactly the path that needs it; regression-guarded).
[...]

PR #179, opened 2026-09-15
fix(docker): close two driver:docker channels that leaked grading material to the agent

CarlesUIPath:
## Summary

Under `driver: docker`, an eval agent could read its own **grading material** as follows: 
1. During execution, the agent has direct access to the grading criteria via /work/input/taks.yaml. In certain cases, the agent trace reasoning compares the expected criteria directly against their own solution. 
2. The skills root mounts the whole repo, and agents can access it. This means an agent could have access to more details about their grading.

This PR closes both channels, adds a lint guard so the skills layout can't recur, and is verified end-to-end with a live adversarial agent. That is, a test which asks the agent to cheat by finding reference files.

The existing reference anti-cheat window #106  (mode-000 on the staged reference copy) did **not** cover either channel: it guards the reference *copies*, while the criteria sat unshielded at `/work/input` and the plugin mount exposed the *originals*.

## The fixes

- **Fix A — delete the staged task definition in-container.** `/work/input` is now writable, and the in-container entry point deletes **both** `task.yaml` and `context.json` immediately after `load_task` (`_scrub_staged_inputs`), before any agent turn on all paths (normal + regrade). Deleting should be a safe op since graders do not read at task.yaml after agent finishes.
- **Fix B — allowlist / default-deny mask of auto-mounted plugin trees.** New `isolation/eval_material.py::mask_dirs`: the plugin root stays mounted `:ro`, but every child dir outside the keep-set (`.claude-plugin` + the manifest-declared skill dirs) is `--tmpfs`-masked. Everything that isn't the skill surface is hidden by default, so an unknown/new eval layout can never leak; `tests/`, references, `node_modules/` are masked for free. This is *stronger* than a chmod window. Graders do not look into the mounted tasks, since these resources are present in task dirs from the previous anti-cheat PR #106.
- **Fix C — CE065 lint.** Flags any `task_id:` YAML or `reference.directory` left readable under an in-repo plugin root (colocated inside a skill dir, or a loose file at the root that a tmpfs can't mask). Reuses `mask_dirs` + `manifest_skill_dirs` as the single source of truth, pinned by an import-identity test.

## Verification

- `make verify`: **5758 passed, 4 skipped** (green; 2 pre-existing env-driven failures only appear when a local `.env` leaks `TASK_DIR`/model — unrelated files, unchanged from main).
- **Live adversarial run** (Claude agent, real container, ordered to peek): came back empty on all channels — `find …/tests…` and `find … reference` → no output; `cat /work/input/task.yaml` and `cat /work/input/context.json` → *No such file or directory*; `verdict.txt: DENIED`; anti-cheat criterion PASS, task SUCCESS.
- Deterministic docker probes + argv/unit tests for `mask_dirs`, `_resolve_mount_path`, nested-plugin dedup, the `skills:"."` stand-down, and `_scrub_staged_inputs`.
- Reviewed by a 3-model pass (Opus/Fable/Sonnet) and a merge+peek verification review — no reachable Critical/High/Medium.

## Known residuals (documented defense-in-depth)

- The reference window's pre-existing gap (root can `chmod` back; full containment needs a non-root uid) is unchanged — out of scope.
- A tmpfs masks a directory, not a single loose file; CE065 covers in-repo plugins but can't reach out-of-repo ones. `skills:"."` stands the mask down (fail-safe, now logged + CE065-flagged).

## Files

`src/`: `isolation/eval_material.py` (new), `isolation/docker_runner.py`, `cli/run_task_internal_command.py`, `agents/_skills.py` (`manifest_skill_dirs` made public/SSOT), `orchestrator.py`, `fs_permissions.py`.
`tests/`: `test_eval_material.py` (new), `test_run_task_internal.py` (new), `test_docker_runner_mounts.py`, `test_custom_lint.py` (CE065).
Docs: `docs/DOCKER_ISOLATION.md`, `.claude/architecture-notes.md`, `CLAUDE.md`.


bai-uipath (commented):
**Fix A is good, Fix B breaks real plugins. Suggest landing Fix A on its own and reworking the rest.**

- **The mask hides the plugin, not just eval material.** The keep-set is only `.claude-plugin` plus the skill dirs, so `hooks/`, `commands/` and `agents/` (directories Claude Code loads by default) get masked too. On the skills nightly's staged `.plugin-root` (UiPath/skills#3546) that drops the plugin's hooks and commands and hides `tests/scripts/stage-preview-sdk-workspace.sh`, which the nightly pre_run calls, so the eval grades a different plugin than the one that ships. Fix before merge: keep the full plugin surface (default component dirs plus every manifest-declared path), or replace the mask with a preflight that errors when a mounted plugin root contains `task_id:` YAMLs.

- **CE065 checks nothing real.** It only walks this repo's `tasks/`, where no task resolves to a plugin root, and never sees the suites where this layout actually lives. Fix: cut it, or fold it into the runtime preflight above so it covers every suite.

- **Needs a rebase, and the comments won't pass main's new prose gate.** Several new docstrings and comment blocks exceed the caps, and the same rationale is repeated across about six files. Fix: one section in `.claude/notes/isolation.md` plus a `Rationale:` pointer.

- **Fix A removes one copy of the criteria, not all of them.** The same YAML (and its siblings) is still in `/work/task_dir`, behind a chmod window that gemini-3.8-flash bypassed in the BPMN sweep. Not a blocker, but the comments saying the agent "cannot read its criteria" should be scoped to `/work/input`.

PR #181, opened 2026-09-16
feat: harness contract and plugin SPI (plan 1 of 3)

uipreliga:
## Summary

Plan 1 of 3 of the harness target architecture: every agent declares what it honors of the shared agent config, and a task that asks for something the chosen harness cannot do is rejected at resolution, so `coder-eval plan` fails before any run.

- **`HarnessContract` per agent class.** Each agent declares which of `system_prompt`, `plugins`, `permission_mode`, `allowed_tools` and `disallowed_tools` it honors, the `permission_mode` values it honors, and whether it supports cooperative stop. Registration fails without a contract. The ClassVars `supports_cooperative_stop`, `supports_cost_log_tags` and `system_prompt_semantics` are gone.
- **A closed tool vocabulary.** `CANONICAL_TOOL_NAMES` (14 names, checked against the pinned Claude Code CLI) and a per-harness `ToolNameMap` that must map every name. Names are checked at resolution, with a did-you-mean hint.
- **Tool restrictions honored natively.** Pi: `--tools` / `--no-tools` / `--exclude-tools`. OpenCode: `permission` rules and an `instructions` file for the system prompt. Antigravity: SDK tool-call policies. `permission_mode: plan` means read-only on all three. The Codex tool forward did nothing (`enabled_tools` is an MCP-server key) and is deleted.
- **Per-kind defaults.** An experiment's `defaults.agent.by_type.<kind>` applies only when the final agent kind matches, below the task. Claude-only defaults in `experiments/default.yaml` moved there.
- **Resolution-time check.** `validate_harness_contract` raises `HarnessContractError`, a `TaskResolutionError`. `plan`, `run` and `export` stop on it. Every "set but NOT enforced" warning is deleted.
- **Plugin surface.** `coder_eval.spi` with `SPI_VERSION = 1`. `cost_log_tags` is a base-constructor kwarg on `Agent`.
- **Guardrails.** Lint rule CE068 keeps concrete agent kinds out of `orchestration/`, `streaming/` and `timing.py`. `make parity-table` generates the contract and tool tables in `docs/agents/HARNESS_PARITY.md`, and CE069 fails on drift. `tests/test_harness_conformance.py` requires one offline probe per enforced cell and per declared permission mode, and checks that a system prompt never rides the user turn.
- **Bugs fixed from review.** Pi and OpenCode admitted one turn over `max_turns`. Antigravity re-pulled the step stream on any `RuntimeError`, and swallowed a failed teardown silently. The early-stop ceiling could divide by zero.

## Breaking changes: YAML that loads today is now rejected

- A Codex task that sets `permission_mode`, `allowed_tools` or `disallowed_tools`.
- `permission_mode: acceptEdits` or `default` on Pi, OpenCode or Antigravity (they honor `plan` and `bypassPermissions`). A task-level `acceptEdits` run with `--type pi` is rejected too; put it under `by_type.claude-code`.
- A tool name outside the canonical set. Claude Code still accepts `mcp__…` names and rules such as `Bash(git status:*)`.
- `plugins`, `system_prompt`, `permission_mode` or a tool list on a `type: none` task.
- `experiments/default.yaml` now applies `model`, `permission_mode` and `allowed_tools` to Claude Code only, so `--type pi` no longer inherits the Claude model id.
- An empty `allowed_tools: []` restricts nothing on every harness, as on Claude Code.

## Before merge

- [ ] **Delegate (`coder_eval_uipath`)**: its agent fails registration until it declares a `contract` and accepts `cost_log_tags`. Land that change, and move `permission_mode` under `by_type.claude-code` in the RPA pipeline's shared experiment YAML, together with or before this PR.
- [x] **One live run per harness** with `allowed_tools: [Read]`: Pi and OpenCode (Haiku 4.5 via OpenRouter) and Antigravity (`gemini-3.5-flash`; it runs Gemini only) each wrote no file and scored 0, while an unrestricted control run on the same model passed on all three. An extra Antigravity run with only `disallowed_tools: [Bash]` scored 1.0 with no `Bash` call, so its policies are selective.

Open follow-ups (simulator safeguard, the `acceptEdits` owner question, deferred lint ideas) are tracked in `c/harness-target-architecture-v2.md` and `.claude/harness-candidates.md`.

## Test plan

- [x] `make verify`: 5989 passed, 2 skipped, 93% coverage; ruff, pyright, custom lint (CE000+) and docs budget clean
- [x] `coder-eval plan` over every task: exit 0; `plan -e` over every shipped experiment: exit 0 except the intended `with-plugin` rejection on the agentless task
- [x] `mkdocs build --strict`; new anchors verified in the built HTML
- [x] Docker staging and Harbor export round-trip tests: a reloaded non-Claude task passes the contract check
- [x] Live smoke runs on Pi, OpenCode and Antigravity: restricted runs wrote nothing, unrestricted controls passed (see Before merge)


uipreliga:
Stacked on this PR: #183 (plan 2 of 3, central enforcement) targets `feat/harness-contract`. Merge this one first; #183 then retargets to `main`.

CarlesUIPath:
Nice PR, the joint contract for cross-harness consistency improves eval reliability. I have ran a multi-model review and a few smoke tests on my end. Below are some findings:

### OpenCode: a host's non-tool `allow` is silently flipped to `deny` by rule ordering (Medium)

**What coder-eval emits (deterministic, from `_inject_config_content`):**

    host {external_directory: allow} + task allowed_tools: [Read, Bash]
      -> emitted: {external_directory: allow, "*": deny, doom_loop: allow, read: allow, bash: allow}
      -> the kept host allow sits at index 0, BEFORE "*": deny at index 1

The kept host rule lands before our `"*": deny` because we drop our own copy of it
(`opencode_agent.py:1018-1019`), while the design intent — the "OpenCode applies the LAST
matching rule" comment at `:1014` and `OPENCODE.md:197` — assumes our rules come last. For a
kept non-tool key they don't. The host-**deny** direction is safe only coincidentally (both
entries are `deny`); the ordering is wrong in both cases but only bites when the host wants
`allow`.

**Confirmed live (opencode via Azure Codex `gpt-5.6-luna`).** Read of a file *outside the
project root* via the read tool (bash denied), varying only permission order:

    A  external_directory: deny  (last)                          -> BLOCKED  (confirms the lever)
    B  external_directory: allow (last)                          -> ALLOWED  (allow wins when last)
    C  external_directory: allow BEFORE "*": deny  <-- emitted   -> BLOCKED

B and C differ only in ordering; C is the exact config coder-eval produces for a host that set
`external_directory: allow` plus a task allowlist, and it resolves to DENY. A denied
`external_directory` blocks out-of-sandbox access — where the system-prompt file and skills
under test live — so the host's (and our own default's) intended allow is silently lost.

This matches the two semantics the adapter's own code already assumes: last-match-wins, and
re-allowing `_NON_TOOL_PERMISSIONS` *after* `permission["*"] = "deny"` in `_permission_config`
(`:964-965`) — only necessary if `"*"` matches those non-tool keys, which the live result
confirms it does.

**Fix (verified at the emit layer):** insert kept inherited non-tool rules *after* our entries —
yields `allow`/`deny` correctly in both directions. Worth a unit test pinning the emitted rule
order. Trigger is narrow (host pre-sets `OPEnon-tool allow AND the
task sets an allowlist), hence Medium.

### CodeQL check (currently red) — 3 errors + 1 note

- **3× `py/missing-call-to-init` (error), test-only.** `Agent.__init__` is now concrete
  (takes `cost_log_tags`, sets lifecycle sta `__init__` without
  `super().__init__(...)` are flagged: `tests/fixtures/mock_agent.py:15` (`__init__` at :30),
  `tests/fixtures/text_stub_agent.py:17` (`_
  `tests/test_simulation_integration.py:106` (`_ExplodingAgent.__init__` at :111). Add the
  `super().__init__(...)` call to each (theyontract = stub_contract()`;
  the init call was missed).
- **1× `py/unused-global-variable` (note) — ALIASES`
  (`models/enums.py:148`) is used in `models/harness_contract.py` (import + `from_inverse`
  alias expansion); the query only reasons wfe to dismiss.

### Low / informational (no change required,

- **`disallowed_tools: ["Task"]` on claude-cium; verification
  brought it to Low. The claude-code adapter forwards `disallowed_tools` raw
  (`claude_code_agent.py:1179`, `:1199`) andIASES` itself (identity
  map at `:709`), unlike the `from_inverse` harnesses — so it relies on the CLI resolving the
  alias. Tested the CLI directly via the modtool list:

      control                 -> subagent ad
      --disallowedTools Agent -> `Task` removed
      --disallowedTools Task  -> `Task` remo

  The pinned CLI resolves the alias in both . Residual concern is
  only that this rests on undocumented CLI behavior the adapter neither owns nor tests — a
  conformance probe asserting both spellingsch a future regression.
  _(CLI 2.1.197; PR pins 2.1.216.)_
- OpenCode's `edit` permission key couples Wirections
  (documented CLI coarseness).
- Simulator `_SIMULATOR_DISALLOWED_TOOLS` liharmless given the
  result above, but list both spellings for clarity.
- `command_executed` matches `tool_name` witle `judge_context` /
  `summaries` accept both spellings — inconsistent; normalize at the one matching seam.
- Pi `plan` + an allowlist collapses to `--nocumented divergence
  from claude-code `plan`.
- `plan` validates merge layers 1–4 only (skexport/harbor still
  catch a `-D` violation, so not a correctness hole.

